PR #2, opened 2019-09-01
Revamped Listener View, Chat View

rauchp:
listener_view_new


EphemeralX:
Looks good, thanks for the refactor! Nice comments, they were helpful.  We can probably cut back on some, but don't have to worry about that right now.

Just some nits and suggestions.  For the UI, the 3 parts (header, broadcaster avatars, and follow button w/ count texts) looks a little asymmetrical?  Also we should circle back with Alex and Cooper on whether we should put the follow button and text above the broadcaster avatars or not (could potentially be confusing if the follow is perceived to be associated with the broadcaster avatar right above it).

Note: Instead of returning empty View, we should be able to just return null right?

rauchp:
Wow, I didn't even notice but it was definitely off-center, thanks for catching that! I fixed it in the latest commit. We should definitely circle back about where the Follow  + viewer count + sub count goes, what I added was just a temp fix. Plus, we need a button for people to subscribe as well, right? The UI already felt like it was getting crowded with just these things, so we're going to have to get crafty. Especially if we want to keep the chat as the focus here.

PR #4, opened 2019-09-03
Create broadcasts graphql

EphemeralX:
Changes
* Add RN UUID Generator library as dependency
* Add GraphQL Mutation call to createBroadcast in DynamoDB whenever a user starts a broadcast (verified in AppSync and DynamoDB, actual resolver implementation is in AWS AppSync Console)
* Add AsyncStorage logic and naming format consistency for USER_INFO and PROFILE_PIC and PUSH_TOKEN
* Register FCM Push Token when initializing session in AuthLoadingScreen
* VSCode IDE automatically formatted a couple files to it's current default settings.
* Fixed NewChatModal bug when creating a channel with users; changed back to callback instead of async/await because SendBird's functions weren't designed for that format (SendBird API rather than JS language feature)


civilapp/current/CurrentlyPlayingScreen.js
```diff
@@ -111,8 +121,32 @@ export default class CurrentlyPlayingScreen extends Component {
         console.log('Setting broadcastTitle in parent:', broadcastTitle)
         try {
           const channel = await sbCreateOpenChannel(broadcastTitle)
-          // TODO_PEDRO: Need to store and save channel URL to DB here.
           console.log('Open Channel created:', channel)
+          // TODO_PEDRO: Need to store and save channel URL to DB here.
+          // TODO: Create DynamoDB entry for broadcast as a whole, also trigger push notification to followers
```
rauchp: these todos aren't necessary anymore I believe

rauchp:
Looks good, but seeing createdAt and isLive has me wondering how channel creation is going to work past initial generation. 

Assuming when isLive is set to off (ideally it would be ~5 minutes after host leaves his channel, to account for app crashes and such.), we kick the other users from channel.

1. How does a host rejoin his channel within that timeframe (as opposed to creating a new channel)? If an app crashes and a user loses his audience he's gonna be pissed, so we should make sure that isn't the default.

2. Do we have a plan to clear these broadcasts in DB after they're no longer live?

PR #5, opened 2019-09-05
Broadcaster Screen Revamp

rauchp:
1_pr
2_pr
3_pr


civilapp/current/common/HostView.js
```diff
@@ -27,7 +29,11 @@ const HostView = props => {
         style={{
           height: width / PROFILE_SIZE_MULTIPLIER,
           width: width / PROFILE_SIZE_MULTIPLIER,
-          borderColor: profile.muted ? GREY_MUTED : GREEN_ACTIVE,
+          borderColor: profile.muted
```
EphemeralX: We should take these nested tertiary logic out and write as if-else and assign it to a variable for better readability. (Goes for the other one(s) in the codebase as well)

EphemeralX: Or use tertiary but have the nested tertiary part in it's own variable.  Goal is to avoid nested tertiary for readability.

civilapp/current/UserModal.js
```diff
@@ -69,7 +210,7 @@ const UserModal = props => {
         <View
           style={[
             styles.hostsAndMicViewStyle,
-            { marginTop: props.platform === "ios" ? 115 : 80 }
+            { marginTop: props.platform === "ios" ? 115 : 115 }
```
EphemeralX: nit: can remove check if values are going to remain the same

civilapp/current/BroadcasterView.js
```diff
@@ -79,26 +109,15 @@ class BroadcasterView extends React.Component {
       profilePic = await Storage.get("profilePic.jpg");
     }
 
-    const { broadcastName, navigation } = this.props;
+    const { broadcastTitle, navigation } = this.props;
+    console.log("perdon?", this.props);
```
EphemeralX: We should remove these debug statements.

civilapp/current/UserModal.js
```diff
@@ -1,13 +1,131 @@
...
+} from "react-native";
 import AddUserView from "./common/AddUserView.js";
 import HostView from "./common/HostView.js";
 import { Icon } from "native-base";
 import { MIC } from "./common/Constants.js";
-const UserAddMenu = props => {
+
+const UserRequest = props => {
+  const { width } = props;
+  console.log("queque", props);
```
EphemeralX: Should remove this

rauchp:
Done, fixed changes.

PR #7, opened 2019-09-09
Redid Auth

rauchp:
1_a

2_a

3_a


rauchp:
not a fan of having to import react-native-phone-input manually, but that library was so broken that my version looks nothing at all like the original version. Thinking maybe I should git pull and fork a version just in case..

rauchp:
Simulator Screen Shot - iPhone X - 2019-09-10 at 00 12 41

Simulator Screen Shot - iPhone X - 2019-09-10 at 00 11 28

rauchp:
> Is there any reason we shouldn't use arrow functions for all functions? It also eliminates the need to explicitly call bind ever.

It's a style thing. For some of these functions that are integral parts of the classes, we should be making them regular functions and binding in constructor (which I got lazy and haven't done yet for these), at least that's how React ideally  wants us to set up class components. Since I had to remake these classes I just instinctively  made functions as class functions, also makes classes more legible too I guess. Up to you which one we stick with.



> Why set picture to 1? Looks like you're using this as a bool flag, can we make it more explicit if that's the case? (boolean imageSelected for instance)

Right, I was thinking of setting state to picture itself, just got sleepy last night, will do. 

> For style consistency, is it cool if we avoid _ prefixed naming?

Whoops yeah, will do

EphemeralX:
I believe normal functions and binding is the old way of doing things; it seems the only con of using arrow functions for everything is potentially losing performance optimization on PureComponents: https://frontarm.com/james-k-nelson/when-to-use-arrow-functions/
Can we stick to arrow functions for all functions for consistency for now?

PR #8, opened 2019-09-16
Error component is displayed when user enters empty input for creating broadcast + sports api layer

carteryu:
1. when user enters empty input when prompted for broadcast title, they are unable to proceed

2. made a file in sports folder for when we need to fetch sports data later. looks like most of them are get requests so extending this library is just a matter of C+P.


carteryu:
> single quotes it is

lmfao my dam autoformatter, ill figure out how to disable it later

civilapp/sports/utils.js
```diff
@@ -0,0 +1,54 @@
...
+        },
+      },
+    )
+
+    if (res.status === 401) {
+      console.log('401 error in getAllActiveTeams()')
+    }
+    data = await res.json()
+    console.log(data)
+    return await res.json()
```
EphemeralX: return data

civilapp/sports/utils.js
```diff
@@ -0,0 +1,54 @@
...
```
EphemeralX: return data

PR #9, opened 2019-09-16
Fixed broadcaster with appsync and graphql input changes.

rauchp:


civilapp/current/BroadcasterView.js
```diff
@@ -112,7 +112,8 @@ class BroadcasterView extends React.Component {
       profilePic = await Storage.get(userId + '_profilePic.jpg')
     }
 
-    const { broadcastInfo, broadcastTitle, navigation } = this.props
+    var { broadcastInfo, broadcastTitle, navigation } = this.props
```
EphemeralX: let

rauchp: you modified broadcastInfo later so it was giving errors for modifying a non-mutable object

PR #10, opened 2019-09-17
GraphQL Subscriptions

EphemeralX:
Listeners and Broadcasters each need to listen to a set of events and render/re-render components based on these events.

Example use cases:

- joinBroadcast(...): | Client subscribes to joinBroadcast upon joining the channel to be notified when other users join broadcast
- leaveBroadcast(...) | Client subscribes to leaveBroadcast upon joining the channel to be notified when other users leave broadcast(?)
- requestToJoin(...) | Client subscribes to requestToJoin when they create a broadcast as a host to be notified when other users request to join the broadcast.
- cancelRequestToJoin(...) | Client subscribes to cancelRequestToJoin when they create a broadcast as a host to be notified when other users cancel their request to join the broadcast
- approveRequestToJoin(...): | Client subscribes to approveRequestToJoin when they request to join a broadcast as a listener, to be notified when the host approves their request.
- denyRequestToJoin(...) | Client subscribes to denyRequestToJoin when they request to join a broadcast as a listener, to be notified when the host denies their request.

Tested with iOS app entering a channel, Android app entering the same channel.  Verified in the logs that an event was triggered and logged the user joining.  We can use this data to set state to render new information.

GraphQL Subscriptions uses websockets behind the scenes between client and server.


civilapp/current/BroadcasterView.js
```diff
@@ -114,7 +114,7 @@ class BroadcasterView extends React.Component {
 
     let { broadcastInfo, broadcastTitle, navigation } = this.props
 
-    const uid = navigation.getParam('uid', userPreview.id) //Math.floor(Math.random() * 100)) //TODO: Use user id once we start creating users in DB
+    const uid = navigation.getParam('uid', Math.floor(Math.random() * 100)) //TODO: userId is UUID string, but Agora wants an int; find a way to cleanly consolidate the 2
```
carteryu: i could be misunderstanding what youre trying to accomplish but once we create user id in DB `Math.floor(Math.random() * 100)` makes it non-unique?

civilapp/current/ListenerView.js
```diff
@@ -134,6 +135,46 @@ class ListenerView extends React.Component {
...
+    });
+
+    const onSetBroadcastStatusSubscription = API.graphql(
+      graphqlOperation(subscriptions.onSetBroadcastStatus, { id: broadcastId })
+    ).subscribe({
+      next: (data) => console.log("Listener onSetBroadcastStatus:", data),
+      error: (error) => console.log("Subscription error:", error)
+    });
+
+    this.setState({
```
carteryu: nit: 

just curious / for my learning: any advantage or reason as to why you don't just have 4 state variables:

        onJoinBroadcastSubscription,
        onLeaveBroadcastSubscription,
        onSetBroadcastTitleSubscription,
        onSetBroadcastStatusSubscription

and instead consolidate into a subscriptions array?

civilapp/current/BroadcasterView.js
EphemeralX: Yeah, this is a temporary hack because Agora wants an int for 'UID' but we store UUIDs as strings.  Going to have to put more thought into a way to convert the UUID to int format (and back if we ever need to).

civilapp/current/ListenerView.js
EphemeralX: I add these into `this.state.subscriptions` which is an array so I can just traverse it and close them all easier in code.  Seems cleaner to me and I think minimizing the number of state variables would be nice.

PR #12, opened 2019-09-19
Added Drawyer Navigation, fixed Settings Screen.

rauchp:
Simulator Screen Shot - iPhone X - 2019-09-19 at 15 37 37

Simulator Screen Shot - iPhone X - 2019-09-19 at 16 15 13

Got it working with Profile routing!


civilapp/AppNavigator.js
```diff
@@ -58,6 +60,21 @@ const AppStack = createStackNavigator(
   },
    stackSetting,
 )
+
+const AppDrawyer = createDrawerNavigator(
```
EphemeralX: nit: Drawer

civilapp/AppNavigator.js
```diff
@@ -58,6 +60,21 @@ const AppStack = createStackNavigator(
...
+  { Civil: AppStack,
```
EphemeralX: nit: newline, line 69 also.

btw, why do we have Settings and 'Log Out' both link to the SettingsScreen?

civilapp/AppNavigator.js
```diff
@@ -76,11 +93,11 @@ export default createAppContainer(
   createSwitchNavigator(
     {
       AuthLoading: AuthLoadingScreen,
-      App: AppStack,
+      App: AppDrawyer,
```
EphemeralX: btw, why do you do Drawyer instead of Drawer?

civilapp/components/CustomDrawer.js
```diff
@@ -0,0 +1,81 @@
...
+  StyleSheet,
+  TouchableOpacity,
+  AsyncStorage
+} from 'react-native';
+import { withNavigation } from 'react-navigation'
+import { DrawerNavigatorItems } from 'react-navigation-drawer';
+
+const backgroundColor = '#1C1C1C';
+
+
```
EphemeralX: nit: unneeded newline

civilapp/components/CustomDrawer.js
```diff
@@ -0,0 +1,81 @@
...
+class CustomDrawerContentComponent extends Component {
```
EphemeralX: nit: add new lines between functions and variable declarations on the class level

civilapp/components/CustomDrawer.js
```diff
@@ -0,0 +1,81 @@
...
+  state = {
+    userId: '',
+    username: '',
+    profileImageUrl: 'https://icon-library.net/images/default-user-icon/default-user-icon-4.jpg' // need default to not provide '' to uri
+  }
+  componentDidMount = async () => { 
+    let userPreview = await AsyncStorage.getItem('USER_PREVIEW');
+    userPreview = JSON.parse(userPreview);
+    const userId = userPreview.id;
```
EphemeralX: Why not remove this line and add it to `const { id, username, profileImageUrl } = userPreview` ?  Since the method ends there and userId isn't used elsewhere within it's scope

civilapp/components/CustomDrawer.js
```diff
@@ -0,0 +1,81 @@
...
+          <DrawerNavigatorItems {...this.props} />
+        </SafeAreaView>
+      </ScrollView>
+    )
+  }
+}
+const ProfileHeader = props => (
+  <TouchableOpacity
+    onPress={props.onPress}
+    style={{
```
EphemeralX: nit: styling this long can be in the stylesheet definition imo

civilapp/components/CustomHeader.js
```diff
@@ -32,7 +32,7 @@ class CustomHeader extends Component {
         <Header style={styles.headerStyle}>
           <Left style={{ flex: 1 }}>
             <Button transparent onPress={this.goBack}>
```
EphemeralX: If we're changing the icon to Menu, we probably want to change the onPress to something other than `this.goBack

civilapp/components/CustomHeader.js
```diff
@@ -43,7 +43,7 @@ class CustomHeader extends Component {
               transparent
               onPress={() => this.props.navigation.navigate('Settings')}
```
EphemeralX: We'll have to change this too

civilapp/settings/SettingsScreen.js
```diff
@@ -1,23 +1,58 @@
...
+                    onPress={async () => {
+                      if (data == 'Log Out') {
+                        const data = await Auth.signOut()
+                        console.log(data)
+                        sbDisconnect()
+                        this.props.navigation.navigate('SignIn')
+                      } else {
 
+                        this.props.navigation.navigate(data)
+
```
EphemeralX: nit: whitespace

civilapp/settings/SettingsScreen.js
```diff
@@ -1,23 +1,58 @@
...
+        <Container style={{ backgroundColor: '#161718' }}>
+          <Content contentContainerStyle={{ backgroundColor: '#161718' }}>
+            <List
+              keyExtractor={(item, index) => index.toString()}
+              dataArray={routes}
+              renderRow={data => {
+                return (
+                  <ListItem
+                    button
+                    onPress={async () => {
```
EphemeralX: nit: in favor of moving this to a function outside of render.

civilapp/components/CustomHeader.js
rauchp: Saving this change for when I make search

civilapp/AppNavigator.js
```diff
@@ -60,11 +61,10 @@ const AppStack = createStackNavigator(
   },
    stackSetting,
 )
-
-const AppDrawyer = createDrawerNavigator(
+const AppDrawer = createDrawerNavigator(
   { Civil: AppStack,
     Settings: SettingsScreen,
-    'Log Out': SettingsScreen,
+    'Log Out': LogoutScreen,
```
EphemeralX: Can we do something like `LogOut` instead of 'Log Out' here?

civilapp/settings/LogoutScreen.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react'
+import { View } from 'react-native'
+import { withNavigation } from 'react-navigation'
+import { Auth } from 'aws-amplify'
+import { sbDisconnect } from '../ChatUtils'
+
+class LogoutScreen extends Component {
```
EphemeralX: Oh you might've misunderstood my previous comment here.  I don't think we should have a LogoutScreen since logout is just a function that takes you back to sign in after cleaning up auth (Auth.logOut, sbDisconnect, etc), rather than a screen.  I think adding a comment to why `LogOut: SettingScreen` above is fine.

PR #13, opened 2019-09-20
Auth flow improvements n shit

carteryu:
1. now checks if phone # input is valid
2. checks if confirmation code is legit
3. checks if username is taken. if so, user cant proceed


rauchp:
what happens if I put in a wrong code?

carteryu:
> what happens if I put in a wrong code? Just the alert right?

Yep we're changing the copy and everything later, just making sure basic logic is functional :P

civilapp/auth/SignInScreen.js
```diff
@@ -86,13 +87,23 @@ export default class SignInScreen extends Component {
...
-      const user = await Auth.signIn(phoneNumber)
-      console.log('Found user info: ', user)
-      //user.challengeParam.USERNAME is user's cognito UUID
-      const userId = user.challengeParam.USERNAME
-      this.setState({ user, userId: userId, showLoading: false, showConfirmationForm: true, newUser: false })
+      if (!this.isValidPhoneNumber(this.state.phoneNumber)) {
+        this.setState({ showErrorComponent: true })
+      } else {
+        this.setState({ showLoading: true })
+        phoneNumber = `+${country_code}` + phoneNumber
```
EphemeralX: nit: Can we update this to camelCase to be consistent with our variable naming?  I know in the actual country picker there is 1 place where it has to be snake_case, which can be the exception.

civilapp/auth/SignInScreen.js
```diff
@@ -106,20 +117,18 @@ export default class SignInScreen extends Component {
...
+      await Auth.sendCustomChallengeAnswer(user, authenticationCode)
+    } catch {
+      console.log('Unable to send custom challenge')
+    }
+    try {
+      await Auth.currentSession()
+      const nextRoute = newUser ? 'UsernameSetup' : 'AuthLoading'
+      this.props.navigation.navigate(nextRoute, { user: this.state.user })
+    } catch {
+      console.log('Apparently the user did not enter the right code')
```
EphemeralX: nit: Can remove this since we have the alert.

civilapp/auth/SignInScreen.js
```diff
@@ -134,8 +143,18 @@ export default class SignInScreen extends Component {
     this.setState({ keyboardVisible: true })
   }
 
+  //For now, only checks if # input is 10 digits
+  //We can use a phone validator library in the future or expand upon this method
+  isValidPhoneNumber = str => {
+    //TODO: this is a more thorough regex search, we can do something like this when requirements are more clear
```
EphemeralX: As mentioned offline, we can do this for now, but we'll ultimately want to use a more elaborate phone number validator that supports international numbers, so probably a library?

civilapp/auth/UsernameSetupScreen.js
```diff
@@ -7,99 +7,112 @@ import {
...
       id: this.state.userId,
       username: this.state.username,
       // followers: {},
       // following: {}
     }
-    const newUser = await API.graphql(
-      graphqlOperation(mutations.createUser, {
-        input: userDetails,
+
+    const userVerify = await API.graphql(
```
EphemeralX: super nit: I'd name this `getUserResp` or `userByUsernameResp`, and below write
`const usernameExists = getUserResp.data.userByUsername.items.length > 0`
`if (!usernameExists) { ... }` to make it more verbose.  Sorry, been brainwashed by big tech companies before, just want to instill good habits even when it may seem silly.

civilapp/auth/UsernameSetupScreen.js
```diff
@@ -7,99 +7,112 @@ import {
...
   keyboardWillHide() {
-    this.setState({ keyboardVisible: false });
+    this.setState({ keyboardVisible: false })
   }
 
   confirmUsername = async () => {
     // TODO: logic to save username to DB here
-    console.log("State:", this.state)
+    console.log('State:', this.state)
     const userDetails = {
```
EphemeralX: As mentioned in person, let's move `userDetails` inside the if statement below since it's not used otherwise.

civilapp/auth/SignInScreen.js
carteryu: For sure - I just wanted to get sort of checker in there. IMO this is low priority and i need to spend some time to find a good library (spent only like 10 min so far and no luck) so i'll push this off for our backlog

civilapp/auth/SignInScreen.js
carteryu: well this is coming directly from state which i think u wrote right lol

civilapp/auth/UsernameSetupScreen.js
carteryu: u fking geed

PR #14, opened 2019-09-20
General Deeplink Implementation, tested on iOS emulator and Android phone

EphemeralX:
Basic Deeplink Implementation uses react-navigation to route to different screens with parameters parsed from the URI.

URI is in the format of `civilapp://broadcast/<broadcast-id>` or `civilapp://user/<user-id>`.  

Tested in iOS emulator by typing in the above URIs in safari, confirmed it launched the app and navigated to the correct screen with parameters.
Tested on Android device, from Android Studio passing in a URL in run config that opens the App to the correct screen with parameters.

Next Steps: Investigate how to implement Universal Links for iOS and Android; this requires a lot more work but it allows deeplinks that navigate to the app store when the app is not installed and also remembers the parameters to go to the correct screen.


civilapp/AppNavigator.js
```diff
@@ -27,15 +27,17 @@ import SocialScreen from './social/SocialScreen'
 
 const stackSetting = { headerMode: 'none', mode: 'modal' };
 
-const CurrentStack = createStackNavigator( 
-  { CurrentMain: CurrentlyPlayingScreen, 
-    Profile: ProfileScreen 
+const CurrentStack = createStackNavigator(
```
rauchp: You probably know this but I removed this StackNavigator, I'm done making routing changes so feel free to merge latest and test with that.

civilapp/social/ChatView.js
```diff
@@ -47,13 +47,11 @@ class ChatView extends Component {
   //TODO: fix, this always runs because there is no this.state in a static method
   static getDerivedStateFromProps(props, prevState) {
     const { messageList } = prevState
-    // if (messageList !== this.state.messageList) {
```
carteryu: I know this was causing issues with chatview rendering but make sure that by removing this it doesnt break any existing functionality with displaying message list

PR #16, opened 2019-09-21
Social chat screens

rauchp:
Simulator Screen Shot - iPhone X - 2019-09-20 at 20 06 16

And I changed routing again, turns out one of the stacks was unnecessary. Code looks a lot simpler now.


civilapp/social/SocialScreen.js
```diff
@@ -1,69 +1,32 @@
...
+import { StyleSheet, ImageBackground, View } from 'react-native'
+import CustomHeader from '../components/CustomHeader'
 import OpenChatsView from './OpenChatsView'
 
 export default class SocialScreen extends React.Component {
   constructor(props) {
     super(props)
   }
 
+  searchUser = () => {
```
EphemeralX: I know it's not in scope for this, but just FYI, we want to have full user search, like Instagram for example, beyond just browsing friend list here.

PR #18, opened 2019-09-23
Basic amplitude logging + sports data api features

carteryu:
1. Basic amplitude logging implemented (i.e. FTUE events are fired off and collected into dashboard. Let me know if you dont have access)
2. Since we know we will need live sports data, I made some more ready to use functions for getting what we need in real time.

TODOs:

1. get a live simulation API key so i can start simulating live events (its offseason for most sports right now)
2. figure out what else we need to log in amplitude.


civilapp/App.js
```diff
@@ -9,13 +9,15 @@ import * as Permissions from 'expo-permissions'
...
 import { pushNotifications } from './push'
+import { amplitude } from './logging'
 
 import Amplify, { Auth } from 'aws-amplify'
 import config from './aws-config' //Uses custom config instead of generated aws-exports.js
 
 useScreens()
 Amplify.configure(config)
 pushNotifications.configure()
+amplitude.init('696969')
```
carteryu: TODO: figure out the best way to assign uuid to a user. Currently thinking of just generating a random one and populating a dynamoDB field to create a relationship.

For now, since we're the only ones testing it, it doens't matter that every event has the same user id in amplitude

civilapp/logging/AmplitudeUtils.js
```diff
@@ -0,0 +1,18 @@
+import amplitude from 'amplitude-js'
```
EphemeralX: Are you using this?

civilapp/logging/index.js
```diff
@@ -0,0 +1,3 @@
+import * as amplitude from './AmplitudeUtils'
```
EphemeralX: nit: can we capitalize here? `Amplitude`

civilapp/package.json
```diff
@@ -9,9 +9,11 @@
   "dependencies": {
     "@react-native-community/push-notification-ios": "^1.0.2",
     "amplify-cli": "^1.0.0",
+    "amplitude-js": "^4.7.0-react-native",
```
EphemeralX: How is this being used?

PR #20, opened 2019-09-24
NFL simulation live data + minor fix to pull to refresh

carteryu:
Very basic implementation of polling / updating live mock NFL data.

Just displaying team score, last play, and time remaining, but we can easily expand more upon this once requirements are more clear and designs are fleshed out.

Also, I forgot to enable onrefresh when I refactored broadcast view the first time. Fixing that here in this PR with a simple fix.


civilapp/browse/NFLSimCard.js
```diff
@@ -0,0 +1,64 @@
...
+    super(props)
+    this.state = {
+      playNumber: 1,
+      currentData: [],
+    }
+  }
+  componentWillMount = async () => {
+    let data = await sims.fetchSimulatedScores(this.state.playNumber)
+    console.log('bp0', data)
+    await this.setState({ currentData: data })
```
EphemeralX: Even though setState is async, it doesn't work with await in JS unless they released a new feature.  If you want to enforce the state updates before executing code, you'll have to pass in a callback: `this.setState({currentData: data}, () => { //code after state update })`

EphemeralX: Also `componentWillMount` is being deprecated; we should do this in `componentDidMount`.

civilapp/browse/NFLSimCard.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+  tick = async () => {
+    let data = await sims.fetchSimulatedScores(this.state.playNumber)
+    this.setState(prevState => ({
```
EphemeralX: Oh didn't know you could reference prevState like that, neat.

civilapp/browse/NFLSimCard.js
```diff
@@ -0,0 +1,64 @@
...
+class NFLSimCard extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      playNumber: 1,
+      currentData: [],
+    }
+  }
+  componentWillMount = async () => {
+    let data = await sims.fetchSimulatedScores(this.state.playNumber)
```
EphemeralX: unless you're going to mutate `data`, use `const` over `let`

civilapp/browse/NFLSimCard.js
```diff
@@ -0,0 +1,64 @@
...
+    console.log('bp0', data)
+    await this.setState({ currentData: data })
+  }
+  tick = async () => {
+    let data = await sims.fetchSimulatedScores(this.state.playNumber)
```
EphemeralX: const

civilapp/browse/NFLSimCard.js
```diff
@@ -0,0 +1,64 @@
...
+        <Text style={{ color: 'white' }}>
+          {currentDataJSON['AwayTeam']} Score: {currentDataJSON['AwayScore']}
+        </Text>
+        <Text style={{ color: 'white' }}>
+          {' '}
+          {currentDataJSON['HomeTeam']} Score: {currentDataJSON['HomeScore']}
+        </Text>
+        <Text style={{ color: 'white' }}>
+          {' '}
+          LastPLay: {currentDataJSON['LastPlay']}
```
EphemeralX: wAnT tO `PLay` smAsH?

civilapp/browse/NFLSimCard.js
```diff
@@ -0,0 +1,64 @@
...
+  componentDidMount() {
+    this.interval = setInterval(() => this.tick(), 5000)
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.interval)
+  }
+
+  render() {
+    let currentDataJSON = this.state.currentData
```
EphemeralX: const

EphemeralX: nit: can we just call it currentData? or `dataJson` maybe?

civilapp/sports/Simulations.js
```diff
@@ -1,6 +1,33 @@
...
+        },
+      },
+    )
+
+    if (res.status === 401) {
+      console.log('401 error in getting simulated scores')
+    }
+    data = await res.json()
+    console.log(data)
+    return data[0]
```
EphemeralX: nit: https://gyazo.com/5c6d90bc288ef910a17650a0f18f103b

civilapp/browse/BrowseScreen.js
```diff
@@ -73,6 +74,7 @@ export default class BrowseScreen extends Component {
           >
             Live Now
           </Text>
+          {/* <NFLSimCard></NFLSimCard> */}
```
rauchp: why tho

civilapp/browse/NFLSimCard.js
```diff
@@ -0,0 +1,75 @@
...
+  // }
+  tick = async () => {
+    const data = await sims.fetchSimulatedScores(this.state.playNumber)
+    this.setState(prevState => ({
+      playNumber: prevState.playNumber + 1,
+      currentData: data,
+    }))
+  }
+  componentDidMount() {
+    this.interval = setInterval(() => this.tick(), 5000)
```
rauchp: is this wired to anything?

civilapp/sports/Simulations.js
```diff
@@ -1,6 +1,34 @@
...
-import { SPORTSDATA_KEY_NBA } from '../settings'
+BTW, looks like only NFL sims are supported for now
+ */
+
+import { SPORTSDATA_KEY_NFL_SIM } from '../settings'
+
+const fetchSimulatedScores = async playNumber => {
+  try {
+    const res = await fetch(
+      `https://api.fantasydata.net/v3/nfl/scores/JSON/SimulatedScores/${playNumber}?key=${SPORTSDATA_KEY_NFL_SIM}`,
```
rauchp: Could we could this url in a map of URL's for the api considering they're per-sport?

carteryu:
@rauchp 

re: "why tho" 
this is just the logic to fetch real time sports data, the point is to use it in the broadcast cards when the designs are done, not as a standalone component :P.

civilapp/browse/BrowseScreen.js
carteryu: the point is to use the logic to pull real time data in the broadcast card itself  not as a standalone component :P. i'll get on that when the designs are released

civilapp/browse/NFLSimCard.js
carteryu: sorry, which part are u referring to? the point of setInterval is to emulate a foreach loop

if it gets too heavy i imagine we can replace this later with probably a webhook (unless u have other suggestions)

civilapp/sports/Simulations.js
carteryu: oh for sure, good feedback! 

definitely keeping it in mind, but for this PR just doing NFL data (since theyre the only real time data stream being offered)

civilapp/browse/NFLSimCard.js
rauchp: this.tick(), what is wired to

carteryu: it refers to the method in my component called tick(), right? :P

PR #21, opened 2019-09-24
Looked through image flow with small edits, also new Room Schema

EphemeralX:
Went thru the image flow code.  In general we want to store local image blob to AsyncStorage's `PROFILE_PIC` key for local user to render (since it's faster than http url).  For all other users, we retrieve the URL from DynamoDB; to ensure the URL is stored in DynamoDB instead of S3 key, we have to do a S3.get request (`Storage.get(<key>)`)which returns a publicly accessible URL to store in DB.

After lots of product discussion, created the initial version of Room schema which mirrors Broadcast model but will be persistent rather than ephemeral.  Renamed and added some fields.  Handing off to Carter to finish so he can get ramped up on the end-to-end GraphQL flow we're using.


civilapp/ios/Podfile.lock
```diff
@@ -1,6 +1,6 @@
 PODS:
   - AgoraRtcEngine_iOS (2.4.1)
-  - Amplitude-iOS (4.8.0)
+  - Amplitude-iOS (4.6.0)
```
carteryu: nice

EphemeralX: what is this sorcery

PR #25, opened 2019-09-27
Add chat notification badge on footer, polls SendBird

EphemeralX:
I'm not happy with the way I've implemented this so please lemme know if you know of a better way to do this.  Basically I'm not sure if I implemented this in the easiest/cleanest way possible, and also I'm not a fan of polling for info but SendBird readily has this information available...

https://gyazo.com/b56cfc19a5057177aed8c16b19de00df

Next Steps/TODO later: We'll probably want something similar in the actual chat channels in the DM screen.


rauchp:
Wait so there's no event listener we can plug Sendbird to instead?

carteryu:
https://docs.sendbird.com/platform/webhooks

just curious if thats a possibility? ^

rauchp:
SendBird gives us access to onMessageReceived event handler, when the sendbird instance is initialized, we can update poll and run the function onMessageReceived()?

rauchp:
I'm assuming onMessageReceived works in background, but since SB is initialized I'm expecting that to happen, I suppose we can add a console log here to check

https://github.com/outlier-collective/mono/blob/6ab7d169c19aa7a1479ca5a7e04aadfd4c5d10ef/civilapp/social/ChatEventHandler.js#L40

EphemeralX:
> SendBird gives us access to onMessageReceived event handler, when the sendbird instance is initialized, we can update poll and run the function onMessageReceived()?

Oh right, forgot about that; yeah I think there is something I can do there.

EphemeralX:
Did some investigation and thinking on other options.  The extra effort to optimize the performance isn't really needed for MVP so just making a note to revisit later.

Basically the options are:
1. Store an aggregate counter for all onMessageReceived events for all channels as state at the parent level and then pass state down 2 layers as props (App -> CurrentlyPlayingScreen -> BroadcasterView || ListenerView || (SocialScreen -> ChatScreen) -> ChatView).  Currently the chat event handlers are tightly coupled with ChatView and is based on the channel.
2. Migrate to Redux so we can use global state in various layers of the app.
3. Store an aggregate counter from webhook callbacks we set up with API Gateway and Lambda.
4. HTTP GET poll
5. SDK method poll (this solution)

Changing the solution isn't trivial so I suggest we work with this for now.

PR #26, opened 2019-09-29
Fixed a bunch of auth bugs, added seamless auth transition on app crash

rauchp:
A lot of small bugs turned into big problems. Before this, we were creating users 2-3 times per auth flow, and weren't correctly setting them with username, profile etc. On top of this, we weren't setting UserInfo/UserPreview correctly on creation, so that was messing up NavDrawer pretty badly. 

I also went ahead and added some logic conditions to make sure the user can't login without having set up their username or profile pic. I feel more confident about the app's security now. See video below for what it looks like in action.

https://drive.google.com/file/d/1jkue8x6YKYmrPqMUwR0VKaXgLnoFZt4B/view?usp=sharing
we uber now bois


civilapp/auth/ProfilePicSetupScreen.js
```diff
@@ -43,7 +39,10 @@ export default class ProfilePicSetupScreen extends Component {
         noData: true
       },
       async response => {
-        this.setState({ picture: 1 });
+        const picture = response.uri
```
rauchp: s3 still broken, I'm fixing that with a separate commit

civilapp/auth/AuthLoadingScreen.js
```diff
@@ -124,6 +124,7 @@ export default class AuthLoadingScreen extends Component {
   }
 
   persistUserInfo = async (userInfo) => {
+    // What's the point of having both USER_INFO and USER_PREVIEW?
```
EphemeralX: USER_PREVIEW is kinda extraneous but the intent was to only store the info needed to render a user. (username, image, etc).  We could just use USER_INFO and pass in a subset of it around also.

civilapp/auth/UsernameSetupScreen.js
```diff
@@ -66,16 +66,7 @@ export default class UsernameSetupScreen extends Component {
...
-        id: this.state.userId,
-        username: this.state.username,
-      }
-      const newUser = await API.graphql(
-        graphqlOperation(mutations.createUser, {
-          input: userDetails,
-        }),
-      )
-      console.log('New User created:', newUser)
+      await AsyncStorage.setItem('USERNAME', this.state.username) //Set local path just for personal images
```
EphemeralX: Comment here is confusing?

civilapp/auth/ProfilePicSetupScreen.js
EphemeralX: All good, this is just storing into S3 right?  Do you want to store the S3 link or the image binary to AsyncStorage?

civilapp/auth/AuthLoadingScreen.js
```diff
@@ -138,41 +139,52 @@ export default class AuthLoadingScreen extends Component {
...
       } else {
-        console.log('No user found in DB')
+        // If no database record but they're logged in. 
+        // Check for valid fields, if not found send back to auth.
+
+        const profilePic = await AsyncStorage.getItem('PROFILE_PIC')
+        const username = await AsyncStorage.getItem('USERNAME')
+
+        if (!username) {
+          throw new Error('Username Error')
```
EphemeralX: Any reason to use exception handling to navigate instead of just navigating and short-circuiting the function here?

civilapp/auth/UsernameSetupScreen.js
rauchp: yeah it is, I copy-pasted from other line of code that had it, will remove

civilapp/auth/AuthLoadingScreen.js
rauchp: function keeps running after calling navigate(), so I had to do error handling to stop it from making a user and going to screen regardless, I'll see if it's more concise doing some elses inside the loop

EphemeralX: I think you can just call `return` after navigating, but I agree it's a little awkward in syntax since react navigation is asynchronous.

PR #27, opened 2019-09-30
Added Modal for Follower/Listeners. Lots of cleanup.

rauchp:
Simulator Screen Shot - iPhone 11 - 2019-09-29 at 23 33 39

That's what it looks with dummy data. I removed and added hooks to actual listeners and followers. This was my attempt at extending modal logic to support this. It's good to have the list functionality because that'll remain after the redesign. Although there are a lot of UI nitpicks I'd change about this screen right now, there's no use since the redesign probably won't look much like this.


civilapp/current/UserModal.js
```diff
@@ -200,78 +231,124 @@ const CloseButton = props => {
...
+      )
+    })
+  return <View />
+
+}
+
+renderList = props => {
+  const { modalSetting, selectedUsername, followers, listeners, width } = props
+  let dataList
+  if (modalSetting == 'listener') {
```
EphemeralX: nit: We should use enums instead of string checks for modalSetting

rauchp: word

PR #28, opened 2019-09-30
Channel Kick, Apple Dev Setup changes

EphemeralX:
Actual diff is just in https://github.com/outlier-collective/mono/commit/1c0e4948446a62973e1a00f8bb89ad37331153ec

* Wrap up moderation code for kicking, need to call Agora dashboard API for this as our Agora rep suggested; I have a follow-up with him on what it's doing behind the scenes.  We'll verify with multi-client testing this sprint if we need additional work to make this smooth.
* Unknown bug fix for push notifications sending.
* Add apple dev account setting and APNS cert for apple push notifications.
* We can ignore the slightly refactored ChatEventHandler; it's still not being used.


civilapp/profile/ProfileView.js
```diff
@@ -123,7 +123,9 @@ class ProfileView extends React.Component {
         })
       )
       console.log('Updated user:', updatedUser)
-      await AsyncStorage.setItem('PROFILE_PIC', image.uri) //Set local path just for personal images
+
+      await AsyncStorage.setItem('PROFILE_PIC', imageUrl)
```
rauchp: do we know if this works for sure?

PR #29, opened 2019-10-01
Improving Sign Up Flow, use same ImagePicker, slight refactoring

EphemeralX:
Actual diff: 6fc72f8 and 0ca640c

* Refactor PushNotification and Amplitude capitalization
* Go through sign up flow and write to DynamoDB at each step for robustness.
* Use the same image-picker in ProfileView and ProfilePicSetupScreen
* Update image/profilePic to remove the long URL parameter that Storage.get(...) returns - I added a bucket policy to make everything in our bucket's public folder, public read for everyone. (We no longer need to switch to aws-sdk's S3 implementation anymore)


rauchp:
Tested this, thanks for doing this! After we merge, we can use https://github.com/oblador/react-native-progress for rendering progress no? Since we get a progress pct as a callback

PR #31, opened 2019-10-01
Fixed Agora, added host rejoin.

rauchp:
Ok so rejoin is broken now because the create modal opens when it is broadcaster, are you taking care of that, since you said you'd clean up that flow? We can either pass in a nav parameter or maybe check if the broadcast id already exists in DB, if so don't bother with init.


civilapp/browse/BroadcastsView.js
```diff
@@ -76,11 +80,14 @@ class BroadcastsView extends Component {
   }
 
   goToBroadcast = async broadcast => {
+
+
```
EphemeralX: nit: :)

PR #33, opened 2019-10-01
Basic flair implementation

carteryu:
Need to work on UI, but this is the basic implementation for setting flairs. Will dive deep when designs are out.


civilapp/settings/SettingsScreen.js
```diff
@@ -1,37 +1,156 @@
...
+      key: 'Cincinnati_Bengals',
+      link: require('../sports/assets/NFL/Cincinnati_Bengals_logo.png'),
+    },
+    {
+      key: 'Cleveland_Browns',
+      link: require('../sports/assets/NFL/Cleveland_Browns_logo.png'),
+    },
+    {
+      key: 'Cincinnati_Bengals',
+      link: require('../sports/assets/NFL/Cincinnati_Bengals_logo.png'),
```
EphemeralX: Duplicate from line 57

civilapp/settings/SettingsScreen.js
```diff
@@ -1,37 +1,156 @@
...
 class SettingsScreen extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      userId: '',
+      showSuccess: false,
+    }
+  }
+  //TODO: finish populating the data array (prob NBA/Hockey/Soccer leagues too before launch? IDK)
+  data = [
```
EphemeralX: nit: We should put this in a config file imo

EphemeralX: also it looks like you're missing some entries based on the asset files in this commit.

civilapp/settings/SettingsScreen.js
carteryu: Oops

PR #37, opened 2019-10-07
Got RoomView.

rauchp:
Still need to wire up request to join, but we're missing table value for that so I guess it'll have to wait. For some reason UI got a little offset after switching to this, but it's minor (at least on iOS.)

Listeners can't kick people out lol.

Simulator Screen Shot - iPhone 11 - 2019-10-06 at 23 57 10


civilapp/current/CurrentlyPlayingScreen.js
```diff
@@ -24,10 +24,12 @@ export default class CurrentlyPlayingScreen extends Component {
 
   componentDidMount = async () => {
     let userInfo = await AsyncStorage.getItem('USER_PREVIEW')
+    
     userInfo = JSON.parse(userInfo)
     this.userInfo = userInfo
     this.userID = userInfo.id
```
EphemeralX: just for my info, what is `this.userInfo` and `this.userID`? I don't see where they're defined?

EphemeralX: nit: userId

civilapp/current/RoomView.js
```diff
@@ -0,0 +1,383 @@
...
+import { Icon, Button, Text, Toast } from 'native-base'
+import LinearGradient from 'react-native-linear-gradient'
+import { Storage } from 'aws-amplify'
+import { API, graphqlOperation } from 'aws-amplify'
+import RoomViewHelper from './RoomViewHelper'
+import * as mutations from '../src/graphql/mutations'
+
+import AgoraAPI from '../agora/AgoraAPI'
+import ChatView from '../social/ChatView'
+import { MIC, LISTENER, FOLLOWER } from './common/Constants.js'
```
EphemeralX: now that I think of it, we should probably create constants for AsyncStorage keys and reference them, at some point. good idea pedro

civilapp/current/RoomView.js
```diff
@@ -0,0 +1,383 @@
...
+                        textAlign: 'center',
+                        fontSize: 15,
+                    }}
+                >
+                    {props.followerCount} Followers
+      </Text>
+            </TouchableOpacity>
+        </View>
+    )
+}
```
EphemeralX: nit: formatting

civilapp/current/RoomView.js
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    // Unused, should be called when hosts are updated.
+    isHost = (broadcast) => {
+        if (broadcast.host.id == userId)
+            return true
+
+        broadcast.cohosts.find((cohost, _) => {
+            if (cohost.id === userId) return true
+        })
```
EphemeralX: is this check just for safety? If data is correct, the above check will always work right?

civilapp/current/RoomView.js
```diff
@@ -0,0 +1,383 @@
...
+        return false
+    }
+
+    componentWillUnmount() { 
+        if (this.state.subscriptions) {
```
EphemeralX: nit: formatting

civilapp/current/RoomView.js
```diff
@@ -0,0 +1,383 @@
...
+            subscriptions
+        })
+    }
+
+
+    leaveBroadcast = async () => {
+        AgoraAPI.leaveChannel()
+        const { isBroadcaster, userInfo, broadcast } = this.state
+        const { id } = broadcast
+        if (isBroadcaster)
```
EphemeralX: nit: I think it's cleaner to always wrap if statements in { } even if it's a one-liner, unless we put it in the same line but even then...meh.

civilapp/current/RoomView.js
```diff
@@ -0,0 +1,383 @@
...
+    }
+    render() {
+        if (this.state.loading) {
+            return (
+                <View style={styles.center}>
+                    <ActivityIndicator animating={true} />
+                </View>
+            )
+        }
+
```
EphemeralX: nit: formatting, and extra whitespace

civilapp/current/RoomView.js
```diff
@@ -0,0 +1,383 @@
...
+                    blurRadius={this.state.blurRadius}
+                    style={styles.backgroundImageStyle}
+                />
+                <LinearGradient colors={gradientColors} style={styles.gradientStyle}>
+                    <KeyboardAvoidingView
+                        style={{ height: height }}
+                        behavior="padding"
+                        enabled
+                    >
+                        <UserModal
```
EphemeralX: are there any props we can remove from this?  seems like a lot...

civilapp/current/RoomViewHelper.js
```diff
@@ -0,0 +1,133 @@
+import { API, graphqlOperation } from 'aws-amplify'
+import * as mutations from '../src/graphql/mutations'
+import * as subscriptions from '../src/graphql/subscriptions'
+
+
+// called in componentDidMount
```
EphemeralX: nit: which file, where?

civilapp/current/RoomViewHelper.js
```diff
@@ -0,0 +1,133 @@
...
+subscribeToBroadcasterEvents = (broadcastId) => {
```
EphemeralX: We still need to wire the data returned from these subscriptions; if they're in a separate file, how are we going to do that?

civilapp/current/RoomViewHelper.js
```diff
@@ -0,0 +1,133 @@
...
+    //Check if user already requested (locally) and call cancelRequestToJoin instead
+    if (requested) {
+        await API.graphql(graphqlOperation(mutations.cancelRequestToJoin, { id: hostInfo.id, user: JSON.stringify(userInfo) }))
+        this.setState({ requested: false })
+    } else {
+        await API.graphql(graphqlOperation(mutations.requestToJoin, { id: hostInfo.id, user: JSON.stringify(userInfo) }))
+        this.setState({ requested: true })
+    }
+    //TODO: Update UI to reflect the state of request (pending, rejected, approved->broadcasterView)
+}
```
EphemeralX: same question here; we'll need to figure out state management between the functions in this file and the caller.  do we pass in a function from the parent maybe?  or should this just exist outside of this utility class?

civilapp/current/RoomViewHelper.js
```diff
@@ -0,0 +1,133 @@
+import { API, graphqlOperation } from 'aws-amplify'
```
EphemeralX: nit: I think RoomUtils.js is a better name, at least consistent with ChatUtils

civilapp/current/UserModal.js
```diff
@@ -191,16 +191,19 @@ const UserActionMenu = props => {
...
           onPress={() => props.onMutePress()}
         />
+        {props.isBroadcaster ? 
         <MenuButton
           icon={'cancel'}
           text={'Hang Up'}
           onPress={() => props.onHangUpPress()}
-        />
+        /> : null 
+        }
```
EphemeralX: nit: formatting

civilapp/current/RoomView.js
rauchp: It's part of the logic for the transition

civilapp/current/RoomView.js
rauchp: I should've passed in just state :/

considering this'll probably get nuked with redesign I won't worry about it now

PR #42, opened 2019-10-08
Add try-catch to every graphQLOperation, a couple bug fixes after pulling the latest code

EphemeralX:
Ignore the branch name :)

Did a lot of little things and kept appending to the PR since no one was looking at it. :)

* Add try-catch to every graphQL call for better debugging (TODO: Make some of these console.error since we do not want to continue if it happens?)
* Bug fix when creating new broadcasts/rooms because userInfo is not userPreview
* Bug fix with no inputText for GIPHY keyboard (doesn't work for me btw, I have the API key added.)
* Added dependencies to start with swiper, animations, transitions, but haven't done anything with them yet.
* Switched to createMaterialTopNavigator on bottom instead of bottomTabNavigator, since it's that but better; got some kinks with swiping to CurrentlyPlayingScreen to iron out though.
* Fix lingering agora uid Int reference
* Fix white flash/flickering when navigating (not sure it's 100% fixed)
* Fix -count error in APIs
* Add optimization with InteractionManager to load logic after animations complete
* Add screen animations setup (Browse/Current/CreateBroadcast transitions don't work properly yet, but should with the new app direction)
* Add background setting to iOS
* Fix AsyncStorage bug with getting user info to write push token to DynamoDB
* Refactor register push token to SendBird
* Add missing import in CreateBroadcastModal


civilapp/AppNavigator.js
```diff
@@ -26,7 +30,43 @@ import ChatScreen from './social/ChatScreen'
 import NewChatModal from './social/NewChatModal'
 import SocialScreen from './social/SocialScreen'
 
-const stackSetting = { headerMode: 'none', mode: 'modal' }
+//TODO: Test and add more transitions
+const handleCustomTransition = ({ scenes }) => {
+  const prevScene = scenes[scenes.length - 2]
+  const nextScene = scenes[scenes.length - 1]
+  if (prevScene) {
+    console.log("PrevScene:", prevScene.route)
```
rauchp: do we need these / what's the plan for these conditions

civilapp/components/chat/GIPHYBoard.js
```diff
@@ -19,7 +19,8 @@ export default class GIPHYBoard extends Component {
   }
 
   componentDidMount = () => {
-    if (this.props.inputText === '') {
+    const { inputText } = this.props
+    if (!inputText || inputText === '') {
```
rauchp: Isn't this the same?

EphemeralX: This gave me an undefined error, so this should fix it.

civilapp/AppNavigator.js
EphemeralX: I should remove this, but I was using this to debug why some transitions weren't hitting the animations I set.

PR #45, opened 2019-10-10
Add event listeners and logs to debug agora problems, slight fix, now…

EphemeralX:
… able to join channels again but sometimes with error


civilapp/components/CustomFooter.js
```diff
@@ -10,54 +10,83 @@ import {
...
     super(props)
   }
 
-  componentDidMount = async() => {
+  componentDidMount = async () => {
     let userInfo = await AsyncStorage.getItem('USER_INFO')
-    userInfo = JSON.parse(userInfo)
-    const userId = userInfo.id
-    this.setState({ userId })
+    if (userInfo) {
```
rauchp: were you getting problems before with undefined userInfo? If it's not loading, we've got a bigger problem here..

EphemeralX: > Nice, so was the problem init and destroy or permissions?

Permissions.

> were you getting problems before with undefined userInfo? If it's not loading, we've got a bigger problem here..

I added it cause I ran into an error, something to watch out for and investigate further.

PR #49, opened 2019-10-14
Subscriptions

carteryu:
Subscriptions work


civilapp/browse/BroadcastsView.js
```diff
@@ -99,7 +104,8 @@ class BroadcastsView extends Component {
   }
 
   //Using explicit method here instead of library validator in case we want to change how things are done in the future
-  isValidUUID = str => {
+  isValid = str => {
+    if (!str || str === null) return false
```
EphemeralX: `!str` should cover `str === null` case right?

civilapp/current/RoomView.js
```diff
@@ -216,26 +244,106 @@ class RoomView extends React.Component {
...
+    ).subscribe({
+      next: data => {
+        parsedResp = data.value.data.onJoinRoom
+        this.setState(prevState => ({
+          ...prevState.room,
+          room: {
+            ...prevState.room,
+            members: parsedResp.members,
+          },
+        }))
```
EphemeralX: had a hard time reading this, is there a better way to write this?

civilapp/current/RoomView.js
```diff
@@ -106,30 +107,55 @@ class RoomView extends React.Component {
...
       adminInfo: room.admin,
       speakers,
       members,
+      totalOnline,
       isBroadcaster,
       userInfo,
       loading: false,
       showInvites: false,
-      //subscriptions,
+      // subscriptions,
```
EphemeralX: Noticed you're updating state inside the function, which is fine. Just wanna make sure we still want to unsubscribe in componentWillUnmount for cleanup.

PR #50, opened 2019-10-14
bug fix and added request join button.

rauchp:
Simulator Screen Shot - iPhone 11 - 2019-10-14 at 18 42 48

Simulator Screen Shot - iPhone 11 - 2019-10-14 at 18 49 50


civilapp/current/UserModal.js
```diff
@@ -444,6 +445,18 @@ renderUserModal = props => {
   )
 }
 
+
+
+renderRequestInviteView = props => {
```
EphemeralX: renderRequestJoinView?

PR #51, opened 2019-10-15
Request to join + proper rendering

carteryu:
working off of pedros branch with new buttons


civilapp/amplify/backend/api/civilapp/stacks/CustomResources.json
```diff
@@ -475,6 +475,43 @@
         }
       }
     },
+    "RequestToSpeak": {
```
EphemeralX: small nit: add Resolver suffix to be consistent with all the other entries (sorry).  Alternatively we can take it out for all of them.

civilapp/current/RoomView.js
```diff
@@ -322,7 +307,23 @@ class RoomView extends React.Component {
...
+      next: data => {
+        parsedResp = data.value.data.onRequestToSpeak
+        if (!parsedResp) return
+        this.setState(prevState => ({
+          ...prevState.room,
+          room: {
+            ...prevState.room,
+            requestsToSpeak: parsedResp.requestsToSpeak,
+          },
+        }))
```
EphemeralX: yo can you explain to me in english what this is doing? monkey brain here.

PR #52, opened 2019-10-16
refresh fix + render all members of room

carteryu:
1. fixed pull to refresh
2.  everyone in the room is rendered when you click the "everyone in room" button (i imagine in the future we'd want a view to see everyone currently in the room along with all the user privileges i.e. like twitch). for now i just wanted something there so the UI is pretty unpolished


civilapp/browse/BroadcastsView.js
```diff
@@ -46,19 +46,27 @@ class BroadcastsView extends Component {
...
+  onRefresh() {
+    this.setState(
+      {
+        refreshing: true,
+        rooms: [],
+        nextToken: '',
+        moreToLoad: true,
+      },
+      () => {
+        this.getRooms()
```
rauchp: 1. Is not calling await this.getRooms() on purpose?
2. not that it matters much but the setState refreshing can be after the getRooms inside the callback no?

EphemeralX: this.getRooms() doesn't return a promise so it's fine.

carteryu: > not that it matters much but the setState refreshing can be after the getRooms inside the callback no?

no idea bruv i thought it was weird to set state in the 2nd param of set state lol. if i can do it why not xD

PR #55, opened 2019-10-19
NEW! Auth Code and Number Setup 

rauchp:
tested on iOS and android 🤧

- back button sucks, think Carter has a higher quality for some reason
- had to make loading button, so if I could be promoted to figma dev that'd be great


civilapp/auth/SignInScreen.js
```diff
@@ -172,68 +176,143 @@ export default class SignInScreen extends Component {
...
+          style={{ flex: 1 }}
+        >
+          <KeyboardText size={30} value={props.value} />
+        </TouchableHighlight>)
+    }
+    return this.renderKeyboardView(KeyboardKey)
+  }
+  renderKeyboardView = (KeyboardKey) => {
+    return (
+
```
EphemeralX: n

civilapp/auth/common/SuperText.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react'
```
EphemeralX: Good idea on SuperText to wrap this.  nits on spacing throughout this file though, sorry.

carteryu: (▀̿Ĺ̯▀̿ ̿) This is the nit police, You're coming with us. (▀̿Ĺ̯▀̿ ̿)

rauchp: officer please I swear it was my linter

PR #63, opened 2019-10-27
Fixed Touchable Problem on Android.

rauchp:
I unironically spent more time on this than on the Browse Screen. That's showbiz baby

1. Simplified code in LandingScreen (neat trick w/ main view as touchable), this fixed the code on initialization before a user signs in. But the screen was still broken after a user signs in, and restarts the app, hm...

2. I found out one of the dependancies (used for Fluid Nav) is on some dumb shit and blocks JS screen input. That's why it was only happening when on the Fluid Nav Stack. It's a one-liner fix, but I had to fork the repo, remove dependency, and add our own modified dependency. Updating packages in the future just got more fun. I'll probably submit a PR so they can get their shit together.

🤠


civilapp/package.json
```diff
@@ -34,7 +34,7 @@
     "react-native-push-notification": "^3.1.9",
     "react-native-reanimated": "^1.3.0",
     "react-native-safe-area-context": "^0.5.0",
-    "react-native-screens": "^1.0.0-alpha.23",
+    "react-native-screens": "https://github.com/outlier-collective/react-native-screens.git",
```
EphemeralX: dang, baller

PR #72, opened 2019-10-30
Live game stats renders correctly

carteryu:
Game ID is hard coded - it should be passed in as props somehow (prob when user creates room or modifies settings they select a game)

assuming the gameID is valid, this renders real time stats on each "card" in the browse screen


civilapp/sports/LiveGame.js
```diff
@@ -0,0 +1,27 @@
...
+        },
+      },
+    )
+
+    if (res.status === 401) {
+      console.log('401 error in getting simulated scores')
+    }
+    data = await res.json()
+    if (data) {
+      return data.Game
```
rauchp: we should have an else here just in case right?

civilapp/sports/utils/QuarterMap.js
```diff
@@ -0,0 +1,10 @@
+const QuarterMap = {
+  '1': '1st Quarter',
+  '2': '2nd Quarter',
+  '3': '3rd Quarter',
+  '4': '4th Quarter',
+  HALF: 'HALFTIME',
```
rauchp: why no ticks for HALF but for 1,2,3,4?

civilapp/sports/LiveGame.js
```diff
@@ -0,0 +1,27 @@
...
+  try {
+    const res = await fetch(
+      `https://api.sportsdata.io/v3/nba/stats/json/BoxScore/${gameId}?key=${SPORTSDATA_NBA_PROD}`,
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    )
```
EphemeralX: is this sportsdata.io's recommended way to call them? I feel like a GET request can easily sniff out our prod key... just seems kinda not secure to put an API key as a query parameter in a GET, not even POST

civilapp/sports/utils/QuarterMap.js
```diff
@@ -0,0 +1,10 @@
...
+  null: 'FINAL',
```
EphemeralX: feel like we should be explicit here rather than use null as a catch-all key

civilapp/sports/LiveGame.js
```diff
@@ -0,0 +1,27 @@
...
+
+    if (res.status === 401) {
+      console.log('401 error in getting simulated scores')
+    }
```
EphemeralX: should we be checking other error codes? or just not 200?

civilapp/browse/common/SportsCard.js
```diff
@@ -20,147 +21,475 @@ import Card from '../../assets/images/browse/card_header.png'
...
+      teamOneScore: data.HomeTeamScore,
+      teamTwoScore: data.AwayTeamScore,
+      gameQuarter: QuarterMap[data.Quarter],
+      minutesRemaining: data.TimeRemainingMinutes,
+      secondsRemaining: data.TimeRemainingSeconds,
+      gameStatus: data.Status,
+    })
+  }
+  componentDidMount() {
+    this.interval = setInterval(() => this.tick(), 500)
```
EphemeralX: is there a rate limit for their API? (in general I think we should make the call at 1000 ms instead?) - just double check with their API use policy because I'm worried that if we call them every second for each client, if they'd rate limit us.

civilapp/browse/common/SportsCard.js
```diff
@@ -20,147 +21,475 @@ import Card from '../../assets/images/browse/card_header.png'
...
+          style={styles.cardImageStyle}
+          source={Card}
+        />
+        <Image
+          id="AbsoluteNoiseImage"
+          style={{ position: 'absolute' }}
+          source={Noise}
+        />
+        {/* End Absolute */}
+        <View id="EmptySpaceTop" style={{ flex: 0.45, width: '100%' }} />
```
EphemeralX: can we do something like this for the 4 empty View stacking elsewhere in the code (maybe padding/margin percents instead)?  if not, nbd

civilapp/browse/common/SportsCard.js
```diff
@@ -20,147 +21,475 @@ import Card from '../../assets/images/browse/card_header.png'
...
+          style={{
+            marginTop: '4%',
+            width: '90%',
+            flex: 0.2,
+            flexDirection: 'row',
+            justifyContent: 'flex-start',
+            alignItems: 'center',
+          }}
+        >
+          {jerrys.map((jerry, index) => {
```
EphemeralX: for this map and the others, do you think it'd be cleaner to put this in a renderJerry() helper function? (still don't get wtf is a jerry btw)

civilapp/sports/utils/QuarterMap.js
carteryu: dude my fking linter is griefing me

that said - for some reason i think this actually works

civilapp/browse/common/SportsCard.js
```diff
@@ -20,147 +21,475 @@ import Card from '../../assets/images/browse/card_header.png'
...
+    }
+  }
+  tick = async () => {
+    //TODO: this game ID comes from the room creation view when the user sets the field "theme" in the graphql schema
+    //Probably want to pass this in as props
+    gameId = '13957'
+    const data = await LiveGameData.fetchGameData(gameId)
+    this.setState({
+      teamOne: TeamDisplayNameMap[data.HomeTeam],
+      teamTwo: TeamDisplayNameMap[data.AwayTeam],
```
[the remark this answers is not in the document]
carteryu: this is not a 2 item map - pls see teamnamemap.js

civilapp/browse/common/SportsCard.js
```diff
@@ -20,147 +21,475 @@ import Card from '../../assets/images/browse/card_header.png'
...
+    >
+      <ImageBackground
+        imageStyle={{ resizeMode: 'stretch' }}
+        style={[
+          styles.imageTopLevelView,
+          { width: size * 0.875, height: size * 0.925 },
+        ]}
+        source={BorderNoGame}
+      >
+        {renderUpperHalf(false)}
```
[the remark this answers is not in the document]
carteryu: @rauchp pedro theyre after u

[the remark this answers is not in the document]
carteryu: kk

civilapp/sports/LiveGame.js
carteryu: Yea i asked them explicitly what their other clients use and they told me they just fetch using GET. ill dig more into that

civilapp/sports/utils/QuarterMap.js
carteryu: null only happens at default state (cuz i set the state to null in the card).

what do u propose?

civilapp/browse/common/SportsCard.js
carteryu: no rate limit if were paying

civilapp/browse/common/SportsCard.js
carteryu: wait sry not on same page. what would that acheieve?

civilapp/browse/common/SportsCard.js
carteryu: i think pedro likes putting it in render functions and i dont but in this specific case i think i just copied what he wrote lol.

lets just leave as is

civilapp/sports/utils/QuarterMap.js
EphemeralX: I looked this up just now and I guess some people do this.  I was gonna suggest to use FINAL as a key and not allow null keys, but I guess it's nbd.   my concern is mostly based around unintended side effects or unclear expectations with passing a 'null' key.

civilapp/browse/common/SportsCard.js
EphemeralX: was just thinking it'd be nice to be consistent in how we create invisible spacing.  I thought having 4 empty <View/> s is just to create padding/empty space, which seems like the intention of this line (just done in a different way).  not really a big deal, more of it a consistency nit.

PR #73, opened 2019-10-31
Fixed stutter on "next" button appear

rauchp:
feels p smooth now 💦💦💦


rauchp:
cody

Cody getting ready to nit me for putting button in /common instead of ../components/

My reasoning is that we will never use this button beyond these two screens, so rather than have it clog up our components, where I want to store commonly used <Stuff />, we have common be stuff just shared by a single subfolder

civilapp/auth/common/NextButtonComponent.js
```diff
@@ -0,0 +1,59 @@
...
+
+    constructor(props) {
+        super(props)
+        
+        this.props.showButton((shouldShow) => this.setState({ shouldShow }))
+        this.props.showLoading((isLoading) => this.setState({ isLoading }))
+
+        this.state = {
+            isLoading: this.props.loadOnInit ? true : false,
+            shouldShow: this.props.showOnInit ? true : false
```
rauchp: you hate to see the extra props, but this is necessary so on transition it doesn't revert back to "resend code"

EphemeralX: Just for my education, we update the state here from props just to re-render only the button right?  The whole point of passing those empty functions to be defined here is to not re-render the parent components, only this?

civilapp/auth/ConfirmSignInScreen.js
```diff
@@ -31,7 +29,6 @@ export default class ConfirmSignInScreen extends Component {
   state = {
     phoneNumber: '',
```
EphemeralX: If we're using this.phoneNumber, then should we remove it from state?

civilapp/auth/common/NextButtonComponent.js
rauchp: Exactly

civilapp/auth/ConfirmSignInScreen.js
rauchp: You right, I'll remove.

PR #75, opened 2019-11-05
Minor fixes

rauchp:
- Wired "resend code" button
- Removed some dead code
- Added padding to end of Flatlist so you can see whole card after scrolling
- Fixed font on Android

Screenshot_1572906889


civilapp/auth/ConfirmSignInScreen.js
```diff
@@ -95,17 +96,21 @@ export default class ConfirmSignInScreen extends Component {
   }
 
 
-
+  resendCode = async() => {
+    // When
```
carteryu: when wut?

civilapp/auth/ConfirmSignInScreen.js
```diff
@@ -58,9 +58,10 @@ export default class ConfirmSignInScreen extends Component {
   componentDidMount() {
     const userId = this.props.navigation.getParam('userId')
     const user = this.props.navigation.getParam('user')
-    const newUser = this.props.navigation.getParam('newUser')
+    const isNewUser = this.props.navigation.getParam('isNewUser')
     const phoneNumber = this.props.navigation.getParam('phoneNumber')
-    this.setState({ userId, user, newUser, phoneNumber })
+    this.user = user
```
carteryu: sry virgin cs researcher here- whats the advantage of using this.user instead of just modifying a state variable 'user'?

civilapp/landing/LandingScreen.js
```diff
@@ -51,21 +51,15 @@ export default class LandingScreen extends React.Component {
   }
 
   componentDidMount() {
-    if (!this.props.transitionMode) {
```
EphemeralX: Do we not need transitionMode anymore?

civilapp/auth/SignInScreen.js
```diff
@@ -59,7 +59,7 @@ export default class SignIn extends Component {
     try {
       const newUser = await Auth.signUp(params)
       const user = await Auth.signIn(phoneNumber)
-      this.props.navigation.navigate('ConfirmSignIn', { user, userId: newUser.userSub, newUser, phoneNumber })
+      this.props.navigation.navigate('ConfirmSignIn', { user, userId: newUser.userSub, isNewUser: newUser, phoneNumber })
```
EphemeralX: `isNewUser` is being used as a boolean but when we pass it as an object in nav params, it could be potentially misused or confusing if that context isn't remembered.  Should we pass in a boolean instead for clarity? i.e. `isNewUser: newUser ? true : false`

civilapp/landing/LandingScreen.js
rauchp: nope, never ended up using that

civilapp/auth/SignInScreen.js
rauchp: you're right, I didn't realize that, will change before I commit

PR #80, opened 2019-11-07
RIP Jerrys

carteryu:
soyboy PR

1. Avatar selection in FTUE renders at appropriate speed
2. replaced jerry logic with speaker array from dynamo

EDIT:
I'm also squeezing in bugfix to browsescreen in this PR just cuz. Basically scroll to refresh actually works as intended now. little shitter bug took me more time than i care to admit><


civilapp/browse/RoomsView.js
```diff
@@ -50,11 +51,10 @@ class RoomsView extends Component {
     }
     await this.getRooms()
   }
-  onRefresh() {
-    this.setState(
+  onRefresh = async () => {
+    await this.setState(
```
EphemeralX: I don't think this works for setState?

civilapp/browse/RoomsView.js
```diff
@@ -68,14 +68,18 @@ class RoomsView extends Component {
...
+        listRoomParams = {
+          limit: ROOMS_RENDERED,
+          nextToken: this.state.nextToken,
+        }
       }
       const rooms = await API.graphql(
         graphqlOperation(queries.listRooms, listRoomParams),
       )
       const { nextToken, items } = rooms.data.listRooms
+      console.log('gay', nextToken)
```
EphemeralX: rm

civilapp/browse/RoomsView.js
```diff
@@ -68,14 +68,18 @@ class RoomsView extends Component {
   getRooms = async () => {
     try {
       if (!this.state.moreToLoad) return
-      let listRoomParams = { limit: 4 }
-      if (this.state.nextToken) {
-        listRoomParams = { limit: 4, nextToken: this.state.nextToken }
+      let listRoomParams = { limit: ROOMS_RENDERED }
+      if (this.state.nextToken !== '') {
```
EphemeralX: Was this change necessary? Below there is another `if (this.state.nextToken){` btw

PR #85, opened 2019-11-12
Big folder restructure - move all code into /src

EphemeralX:
Moved all code into `src/` folder.  Updated all imports.  Kept assets top-level.  Ran app locally and navigated to all the pages.


rauchp:
no folders for ChatUtils, ChatEventHandler? We could put those in chat right

PR #91, opened 2019-11-16
NEW! RoomView Pt.2

rauchp:
This is a lot of code, so I decided this was a good stopping point. This PR has transformed into a fullstack chad. Tested on both iOS and Android emulators, on-device iOS as well.

My philosophy and thoughts while making RoomView:
 - Agora has great listeners for subscriptions, but RTC can't return a list of speakers,  and it definitely does have a function for listing people that are already muted. My thinking with the RoomView is to use the DB as an oracle for initial state, but nothing more. After you're logged and have subscriptions initialized, you don't rely on the DB until you next go to another room. This means operations such as joining/leaving the channel and muting yourself are reflected in the DB when a user performs them. After testing and changing some things around, I've eliminated a lot of race conditions that lead to weird bugs. Ofc, I don't think this is even 80% bug-free. We need to do a lot of tests, and I think most importantly move around subscription inits to avoid bugs relating to racing. 

I have:
- Only members can chat and use the voice channel. We run DB check and add to DB member array when they join. If you're not a member, I've blocked the text input, and made a shitty grayed out icon for the voice channel.
- Users can enter and leave the voice channel. There is full persistence here, so if I join the speaker channel, and someone joins the room after, they'll see my icon in the speaker slot. If I quit the app or leave the room (or the app crashes), I will disappear from others speaker slots
- You can see who is actively talking. The rendering here needs work. AgoraAPI in this regard works better locally than remotely, so you'll see that it's more sensitive when you talk vs. when others talk.
- You can mute yourself, this is also persistent. If I mute myself and someone joins, they'll see my icon muted, and won't be able to hear me.
- When you click on a talking user, the modal appears displaying either profile or mute. I'd like to extend this to chat as well.
- Redid TextInput at the bottom of the screen. **Right now it conflicts with Cody's footer**, I don't think that's intentional, I'm guessing Cody you want to hide when route ==='Current'?. It has the GIF Board still, and when you click on it the icon changes and you can send your text.
- Redid how chats are displayed. We will have to change some code to get this working for DM's, but we can worry about that when we get there. 
- Did some more retro graphic texts. SuperText is getting filled up, I think I'll split it out in my next PR.

I won't do these in this PR, but the one after, since I already have enough code here:
- Redo GIF View
- Add muting other users locally
- Add Modal Click to text chat. 
- Make 'online' work in realtime, I can do this now, but I'm unsure about how often we want to be pulling this data from Sendbird.
- Understand why GraphQL subscriptions timeout on app background. We really need to make this doesn't happen.
- Wire SportsData, should be straightforward. We need to make a custom clock component tho.

- Fix React Navigation Routing and re-updating screen for RoomView. Right now we use componentDidMount, I'd like to separate that logic to an init function, and call that whenever roomID changes.

Pics:
 

The footer problem that appeared after merging with master:


civilapp/android/app/src/main/AndroidManifest.xml
```diff
@@ -42,7 +42,7 @@
         android:name=".MainActivity"
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
-        android:windowSoftInputMode="adjustNothing"
+        android:windowSoftInputMode="adjustResize"
```
rauchp: had to do it for keyboardAvoidingView on Android. I had to do adjustNothing to get it working before on DM's page.. so we'll see what I do when I get there again. I believe I know why it didn't work before.

civilapp/browse/RoomsView.js
```diff
@@ -254,9 +254,7 @@ class RoomsView extends Component {
               roomPhoto={item.photo ? item.photo : ''}
               //Temp bandaid. We should be passing in the room pic which doesnt yet exist in schema
               //We might even scrap that.
-              hostAvatar={
-                item.admin.profileImageUrl ? item.admin.profileImageUrl : ''
-              }
+              hostAvatar={''}
```
rauchp: This code was breaking master, and it wasn't a DB field issue. Let's not push broken code to master pls

civilapp/components/WoodenBackground.js
```diff
@@ -42,7 +46,7 @@ const styles = {
     width,
     flexDirection: 'column',
     alignItems: 'center',
-    paddingTop: 60, // TODO: adjust for notches
+    paddingTop: 7.5, // TODO: adjust for notches
```
rauchp: ignore the todo, my safeareview fixed notches, will remove comment.

civilapp/current/RoomView.js
```diff
@@ -7,150 +7,253 @@ import {
...
+        const mutedUser = JSON.parse(obj)
+        map[mutedUser.id] = 1;
+        return map;
+      }, {})
+    } else {
+      this.isMutedDict = {}
+    }
+
+    // I need all these to handle self-muting and displaying other's self-muting.
+    AgoraAPI.subscribeAudio((data) => this.handleVolumeIndictaor(data, userInfo))
```
rauchp: 3/4th of this function will be to an init() I can call from didUpdate, for React Nav purposes.

civilapp/RoomAPI.js
```diff
@@ -57,6 +58,28 @@ const leaveRoom = async (id, user) => {
...
 }
 
+const addMember = async (id, user) => {
+  try {
+    const addMemberResp = await API.graphql(
+      graphqlOperation(mutations.addMember, { id, user }),
+    )
+    return addMemberResp.data.addMember
+  } catch (error) {
+    throw new Error('Unable to add as speaker', error)
```
rauchp: need to change to member

civilapp/RoomAPI.js
```diff
@@ -68,10 +91,49 @@ const addSpeaker = async (id, user) => {
...
+}
+
+const addMutedUser = async (id, user) => {
+  try {
+    const addMutedUserResp = await API.graphql(
+      graphqlOperation(mutations.addMutedUser, { id, user }),
+    )
+    return addMutedUserResp.data.addMutedUser
+  } catch (error) {
+    throw new Error('Unable to add as speaker', error)
```
rauchp: need to change as well

civilapp/browse/RoomsView.js
carteryu: how so? only way this breaks is if item.admin doesnt exist - which is always a safe assumption. 

master works and renders just fine for me, what are you running into?

rauchp: yeah mb here, turns out it was some faulty rooms so I cleared them all, will undo this

rauchp: REEEE nvm it wasn't a db thing, this is the error I been getting with this code
 Screen Shot 2019-11-16 at 11 06 11 PM

carteryu:
Thx for doing all the tedious work. Happy to take over sportsdata wiring since I already did it on the browse screen.

gonna ping u with sum bugs

civilapp/current/RoomViewUtils/UserList.js
```diff
@@ -0,0 +1,92 @@
...
+import { getFlair } from '../../sports/utils/FlairUtil'
+// Assets
+import GreenMic from '../../assets/images/room/green_mic.png'
+import RedMic from '../../assets/images/room/red_mic.png'
+import GreyMic from '../../assets/images/room/grey_mic.png'
+import MicOff from '../../assets/images/room/mic-off.png'
+
+const { width } = Dimensions.get('window')
+
+class UserList extends React.Component {
```
EphemeralX: it's kind of confusing to me what this class is supposed to be

civilapp/current/RoomViewUtils/UserList.js
```diff
@@ -0,0 +1,92 @@
...
+            return (
+                <TouchableOpacity onPress={() => muteUser(user)} style={{ marginLeft: -size / 5, padding: size / 5, flexDirection: 'column', alignItems: 'center' }}>
+                    <View style={{ justifyContent: 'center', alignItems: 'center', height: size, width: size, backgroundColor: 'white', borderRadius: size / 2 }}>
+                        <Image style={{ borderWidth: (isSpeaking | isMuted) ? 3.75 : 0, borderColor: color, height: size, width: size, borderRadius: size / 2 }} source={getAvatar(user.profileImageUrl)} />
+                        {isMuted ? <View style={{ justifyContent: 'center', alignItems: 'center', position: 'absolute', height: size / 1.1, width: size / 1.1, borderRadius: size / 2.2, backgroundColor: 'rgba(0,0,0,0.8)' }}>
+                            <Image source={MicOff} />
+                        </View> : null
+                        }
+                    </View>
+                    <Image style={{ left: '50%', top: isSpeaking ? '40%' : '60%', position: 'absolute', height: size / 1.15, width: size / 1.15 }} resizeMode="contain" source={getFlair('LAL')} />
```
EphemeralX: TODO: remove hardcode, wire

civilapp/current/RoomViewUtils/index.js
```diff
@@ -1,5 +1,37 @@
...
+  const onRemoveSpeakerSubscription = await API.graphql(
+    graphqlOperation(subscriptions.onRemoveSpeaker, { id }),
+  ).subscribe({
+    next: data => callback(data),
+    error: error => console.log('Subscription error:', error),
+  })
+  return onRemoveSpeakerSubscription
+}
+
+export { setOnAddSpeakerSub, setOnRemoveSpeakerSub }
```
EphemeralX: maybe this belongs somewhere else? RoomAPI or something else?

rauchp: why?

civilapp/current/RoomViewUtils/UserList.js
rauchp: I guess I could cal it SpeakerList

civilapp/current/RoomView.js
```diff
@@ -7,150 +7,253 @@ import {
...
+        })
+        this.updateUserList('speakers', oldSpeakers)
+        this.speakers = oldSpeakers
+      } else {
+        this.updateUserList('speakers', [])
+        this.speakers = []
+      }
+    })
 
+    let isMember = members.some((rawMember) => {
```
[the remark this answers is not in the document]
rauchp: I literally can't use ```includes``` when I have to parse an object.

civilapp/current/RoomView.js
```diff
@@ -318,44 +382,139 @@ class RoomView extends React.Component {
...
+        <Text style={styles.cancelButtonTextStyle}>Cancel</Text>
+      </TouchableOpacity>
+    )
+
+    // MuteButton Logic 
+    let muteText
+    if (selectedUser.id === userInfo.id) {
+      muteText = isMuted ? 'Unmute' : 'Mute'
+    } else {
+      // TODO: Implement muting others.
```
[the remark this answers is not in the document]
rauchp: It's a little more than that, we need to manage two different mute lists, and probably need to check out persistence, so users can't evading the mute of others just by entering and exiting the speaker channel.

PR #104, opened 2019-11-27
More fixes

carteryu:
1. Fixed all the bugs with profile (except refreshing page when user presses save... need to figure that out tomorrow)
2. Live game data inside roomview
3. CSS stuff here and there


civilapp/src/profile/ProfileEditScreen.js
```diff
@@ -127,8 +131,11 @@ class ProfileEditScreen extends React.Component {
             style={{
               color: '#E25F4F',
               fontFamily: 'Sportsball',
+<<<<<<< HEAD
+=======
               fontSize: 12,
               zIndex: 99,
+>>>>>>> 29f06e556e6ba1cec032aa6bc04a221ec68daedc
```
EphemeralX: I removed when pulling your changes in but don't forget in case you work on top of this more

PR #105, opened 2019-11-27
Pedro bug fixes

rauchp:
from the depths of the costa rican jungle emerges a bug catcher to catch them all

- fixed roomview member update on details page.
- fixed duped key bs for list
- started responsiveness for roomview, need to work with Cody on it
- fixed jumpy isSpeaking render stuff
- fixed user not being speaker on room init


civilapp/src/current/RoomView.js
```diff
@@ -299,6 +301,14 @@ class RoomView extends React.Component {
     }
   }
 
+  componentDidUpdate = async(prevProps, prevState) => {
```
rauchp: Pedro, why are you such a cuck and use componentDidUpdate instead of didMount?

good question, this.updateSpeakerList() wasn't set by the time didMount hit (even though didMount is called post-render), but didUpdate does work, so I used this.

civilapp/src/discover/DiscoverRoomsView.js
```diff
@@ -181,7 +181,7 @@ class DiscoverRoomsView extends Component {
         style={{ flex: 1, width, marginTop: 20 }}
         contentContainerStyle={{ paddingBottom: width / 3.5 }}
         data={this.state.rooms}
-        keyExtractor={(item, index) => item.id}
+        keyExtractor={(item, index) => item.id+''+index}
```
rauchp: the problem was discoverRoomView and the other draft or whatever flattest were using same key, this separates them

civilapp/src/current/RoomView.js
```diff
@@ -299,6 +301,14 @@ class RoomView extends React.Component {
...
+    if (this.props.roomInitialSetup && !this.addedSpeakerInitially) {
```
EphemeralX: do we 100% need both?

rauchp: we dont, I fixed it lol

PR #107, opened 2019-11-29
Misc fixes, updates.

EphemeralX:
* Update deeplinks pre-emptively
* Make back button in RoomView leave the audio channel
* Preload subscribedRooms so Court loads faster (add userId nav param to get rooms in LoadingTransition, pass as preloadedRooms prop)
* Move Agora initiation back to top-level, expecting loading to be faster in RoomView (still need to verify)
* Update `getSpeakerObj` implementation to use as hash map first before calling UserAPI to get the user.  NOTE: This can be further optimized by not passing all the speakerIds every time getSpeakerObj is called, only what's missing.


civilapp/src/App.js
```diff
@@ -30,6 +31,7 @@ We might want to use device id for pre-signUp and find a way to rewire to uuid a
 For now, since we're the only ones testing it, it doens't matter that every event has the same user id in Amplitude
 */
 Amplitude.init('696969')
```
rauchp: nice

civilapp/src/current/RoomView.js
```diff
@@ -273,6 +286,7 @@ class RoomView extends React.Component {
     })
   }
 
+  //TODO: Currently unused
```
rauchp: we need to decide, will going out of focus on this screen (e.g swiping out from here) call LeaveRoom?

PR #110, opened 2019-11-30
general fixes to the app

carteryu:
all the pages now correctly render "default" just chatting text when no game is selected (since u cant select games on create room.. i just never changed it to reflect that until now)

edit: new pushes: 
1. quarter and timer shows .. tested for upcoming, during and after game. there could be 1 or 2 edge cases i havent sat thru a whole game yet to experience shit like halftime

2. removed color b ar
3. log out confirmation shows


civilapp/src/browse/common/SportsCardNew.js
```diff
@@ -134,20 +138,37 @@ class SportsCardNew extends Component {
...
-                    textAlign: 'center',
-                    top: '3.5%',
-                    fontSize: 10,
-                    color: '#78BE20',
-                    fontFamily: 'Sportsball',
-                  }}
-                >
-                  {this.props.roomTeam1} VS {this.props.roomTeam2}
-                </Text>
+                {this.state.gameSelected ? (
```
EphemeralX: can have this check outside of the CSS to define a variable that's either `JUST CHATTING` or ` {this.props.roomTeam1} VS {this.props.roomTeam2}`, then use that variable in the `<Text>` component to make the CSS shorter and cleaner. i.e. `const listenerText = this.state.gameSelected ? `{this.props.roomTeam1} VS {this.props.roomTeam2}` : 'JUST CHATTING'` .... `<Text ...>{listenerText}</Text>`

civilapp/src/current/RoomDetails.js
```diff
@@ -221,36 +226,48 @@ class RoomDetails extends React.Component {
...
-          style={{
-            flexDirection: 'row',
-            alignItems: 'center',
-            width: '100%',
-            textAlign: 'center',
-            justifyContent: 'center',
-          }}
-        >
-          <Text
+        {this.state.gameSelected ? (
```
EphemeralX: Same idea; can do this with styles as well (i.e. define a variable for style and plug it into the component depending on a conditional evaluated outside of the core render CSS code)

civilapp/src/discover/DiscoverRoomsView.js
```diff
@@ -192,22 +203,16 @@ class DiscoverRoomsView extends Component {
...
                 roomChatChannel={item.chatChannelUrl}
-                //How do you get this info from sendbird?
                 unreadMessages={0}
-                roomTitle={item.title ? item.title : 'HORNETSFANZ'}
-                //This should be taken from game Id (I will change this later once we set game id properly in the create room modal or room edit panel)
-                roomTeam1={item.gameOne ? item.gameOne : 'Hornets'}
-                roomTeam2={item.gameTwo ? item.gameTwo : 'Lakers'}
+                roomTitle={item.title ? item.title : 'NO TITLE'}
+                roomTeam1={homeTeam ? TeamDisplayNameMap[homeTeam] : ''}
+                roomTeam2={awayTeam ? TeamDisplayNameMap[awayTeam] : ''}
```
EphemeralX: should we change all references of roomTeam1 and roomTeam2 in the code to homeTeam and awayTeam?

PR #113, opened 2019-12-02
game stats in room now .. works *fingers crossed*

carteryu:
tested for like 10 min while a game was going on a few hours ago. I need to do some edge casing testing (i.e. what happens when halftime or game is cancelled) but im not seeing any issues from all the rooms i visit right now.

Upcoming games and past games seem to work great. Switching games is now a smooth transition


civilapp/src/current/ProfileChangeScreen.js
```diff
@@ -0,0 +1,217 @@
...
+      } else {
+        return { selectedAvatar: key }
+      }
+    })
+  }
+
+  updateAvatar = async avatar => {
+    await AsyncStorage.setItem('AVATAR', avatar)
+    const { userId } = this.state
+    this.props.navigation.navigate('FlairSetup', { userId })
```
EphemeralX: Unless we're gonna merge the 2 screens, we should remove this since we don't want to go to FlairSetup ever from this flow.

civilapp/src/current/ProfileChangeScreen.js
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  render() {
+    return (
+      <RetroBackground>
+        <View style={{ height: height * 0.05 }} />
+        <TouchableOpacity
+          onPress={() => {
+            this.props.navigation.navigate('UsernameSetup', { userId })
```
EphemeralX: Unless we're gonna merge the 2 screens, we should remove htis since we don't want to go to UsernameSetup ever from this flow.

PR #120, opened 2019-12-04
FTUE Massive Update + Misc fixes

EphemeralX:
* Pixel consistent FTUE flow
* Merge ProfileChangeScreen and FlairSelectScreen extra logic into original FTUE versions
* Fix LogOut Modal padding
* Add 2 second delay in AuthLoading to prolong splash screen duration before LandingScreen
* Small OnlineCount updates (not perfect yet)


civilapp/src/auth/FlairSetupScreen.js
```diff
@@ -78,11 +92,31 @@ class FlairSetupScreen extends Component {
...
+  updateFlair = async flair => {
+    const { userId, onSignUpFlow, updateUser, roomId } = this.state
+
+    if (!onSignUpFlow) {
+      if (updateUser) {
+        const { params } = this.props.navigation.state
+        params.saveFlairDraft(flair)
+        this.props.navigation.pop()
+      } else {
+        //TODO: Uncomment when we're ready to save the draft properly without getting overwritten by EditRoomDetails.initScreen()
```
carteryu: just wondering do u have any local changes to push before merging?

civilapp/src/current/EditRoomDetails.js
```diff
@@ -108,6 +108,13 @@ class EditRoomDetails extends React.Component {
     this.setState({ roomBio })
   }
 
+  //TODO-Carter: Save draft of what user changes on this screen, only save when SAVE is pressed.  Currently initScreen is being called on re-focus and it overwrites user selection, we'll have to find a workaround fo rthis.
```
carteryu: i already did the workaround in profile edit!

carteryu: il take care of this just merge when ur ready thx

civilapp/src/auth/FlairSetupScreen.js
```diff
@@ -3,6 +3,7 @@ import {
   StyleSheet,
```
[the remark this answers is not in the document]
EphemeralX: My reasoning is this:
1. Selecting avatar/flair is less likely to change since the functionality is simple, if the UI changes, it should probably change for both for consistency (i.e. there is no good reason to make something like avatar or flair select look different in 2 different places).
2. Discover and Browse are the same for now but they will definitely differentiate more in the future, to my understanding UI included; the logic to get Rooms turned out to be very different also (subject to change though).
3. Realistically the shared code should be refactored further into reusable components anyways, but with the amount of bugs/feature work left, I'll worry about that later/eventually.
4. It's not too hard to debug as the diff in logic between the 2 screens is pretty small and there is an 'onSignUpFlow' boolean to differentiate; it's much more frustrating to make UI edits in 2 places imo (as I've had to do).
5. Less code the better imo, less files to manage; the only caveat is that the files are not where you'd expect but that should be fixed anyways as per point 3.

nbd, just explaining my thinking.

PR #121, opened 2019-12-05
General bugfix

carteryu:
team logo update
improvements all around to profile and profile edit

pretty happy with the profile logic (minus the off by 1 character bug at max character input).


civilapp/src/profile/ProfileEditScreen.js
```diff
@@ -138,18 +160,20 @@ class ProfileEditScreen extends React.Component {
     }
 
     if (!regex.test(text) && text.length < USERNAME_MAX_LENGTH) {
```
EphemeralX: Why don't you just add the `=` to `text.length <= USERNAME_MAX_LENGTH` and fix it?

civilapp/src/profile/ProfileEditScreen.js
```diff
@@ -138,18 +160,20 @@ class ProfileEditScreen extends React.Component {
...
-    const { userId, username, newUsername, profileImageUrl, flair } = this.state
+    const { userId, username, profileImageUrl, flair } = this.state
 
-    //Username check only if the username updated and is different
-    if (newUsername && username !== newUsername) {
-      const existingUserCheck = await UserAPI.getUserByUsername(
-        newUsername,
-      )
+    //This check isn't really necessary since the save button is grayed out when username === initialUsername but it doesn't hurt I guess.
+    if (username !== this.initialUsername) {
```
EphemeralX: I added this before you made your version of the fix; imo we should remove this if we don't need it.  Less code the better.

civilapp/src/profile/ProfileEditScreen.js
```diff
@@ -173,17 +197,27 @@ class ProfileEditScreen extends React.Component {
...
 
   saveFlairDraft = flair => {
     this.setState({
-      flair
+      flair,
+      didChangeFlair: this.initialFlair === flair ? false : true,
     })
   }
-
+  checkSaveButtonEnabled() {
```
EphemeralX: nit: I'd rename this to `canSave()`

civilapp/src/profile/ProfileEditScreen.js
```diff
@@ -367,38 +402,48 @@ class ProfileEditScreen extends React.Component {
...
                       style={{
-                        flex: 0.6,
+                        flex: 0.55,
                         textAlign: 'center',
                         fontSize: 28,
                         fontFamily: UPHEAVAL,
                         color: '#FFFFFF',
+                        //Add a bit of right padding to make up for the fact that theres whitespace in the logo
+                        //Not really sure what the 'perfect' solution is here since every logo has diff amounts of white padding..
+                        //Could be reasonable to just ask the designer/artist to make the horizontal whitespace padding consistent
```
EphemeralX: nit: unneeded

civilapp/src/profile/ProfileScreenNew.js
```diff
@@ -389,35 +389,46 @@ class ProfileScreenNew extends React.Component {
...
+                  }}
+                >
+                  <Image
+                    style={{
+                      flex: 1,
+                      width: undefined,
+                      height: undefined,
+                    }}
+                    source={flair ? getFlair(flair) : null}
+                  ></Image>
```
EphemeralX: nit: `/>` . Also are you *SURE* that width: undefined and height: undefined is necessary? With `<Image>` you only need to specify width and height explicitly if the source is from the URL; otherwise it can take local images dimensions and infer it.

PR #122, opened 2019-12-05
Pixel tweaks

EphemeralX:
* Some Android UI fixes, trying stuff in Agora in attempt to remove background noise
* Pixel tweak and keyboard fix for ProfileEdit
* More pixel tweaks on Profile, EditProfile, CreateRoom, consistency stuff
* Fix EditRoomDetails keyboard blocking text input, make EditRoom->FlairSelect go back to RoomView instead of BrowseScreen, Pixel tweaks for Android


civilapp/ios/civilapp.xcodeproj/project.pbxproj
```diff
@@ -576,7 +576,7 @@
 				CODE_SIGN_ENTITLEMENTS = civilapp/civilappDebug.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 1.0.1.9;
+				CURRENT_PROJECT_VERSION = 1.0.1.10;
```
carteryu: coulda swear ive seen this in the PR at least 2-3 times now lol

civilapp/src/agora/AgoraAPI.js
```diff
@@ -142,21 +142,24 @@ const enterListenerChannel = async (channelName, userId) => {
     await RtcEngine.setEnableSpeakerphone(true)
     await RtcEngine.enableAudioVolumeIndication(500, 3)
 
-    const connectionState = await RtcEngine.getConnectionState()
-    console.log("RtcEngine connectionState:", connectionState)
+    // const connectionState = await RtcEngine.getConnectionState()
```
carteryu: nit - id add a comment as to why its commented out (either in the code or PR) or remove the lines from the file altogether

civilapp/src/auth/FlairSetupScreen.js
```diff
@@ -187,7 +187,7 @@ class FlairSetupScreen extends Component {
                   if (onSignUpFlow) {
                     this.props.navigation.navigate('ProfileSelect', { userId })
                   } else {
-                    this.props.navigation.pop()
+                    this.props.navigation.goBack()
```
carteryu: i cant seem to find the difference between pop() and goback(). both docs say they just take u to the previous page. for my understanding what is the diff?

civilapp/src/browse/CreateRoomScreen.js
```diff
@@ -165,7 +165,9 @@ export default class CreateRoomScreen extends React.Component {
 
 const styles = StyleSheet.create({
   headerView: {
-    marginTop: 20,
+    // marginTop: 20,
```
carteryu: thanks for fixing- if 2% makes sense no need for this. rm

civilapp/src/auth/FlairSetupScreen.js
EphemeralX: Since react-navigation is stack-based, it seems like pop takes you out of the stack and goBack goes to the previous screen on the stack. (When I changed this, I noticed that pop took me back to the Browse Screen when the path was BrowseScreen -> RoomView -> RoomDetails -> EditRoom -> FlairSelect, but goBack took me back to RoomView (which made more sense).  I can see why RoomDetails wouldn't be a target cause it's a drawer but I wasn't sure why it didn't take me back to the EditRoom screen so that's a pending ?.

civilapp/src/agora/AgoraAPI.js
EphemeralX: sure I'll remove them in my next commit; they were there for debugging purposes initially.

civilapp/src/browse/CreateRoomScreen.js
EphemeralX: oh I forgot this one, ty

PR #123, opened 2019-12-05
Made RoomDetails truly responsive to updates.

rauchp:
Added direct communication between RoomDetails ("Header") and its children. Now it's truly responsive. When you update a room it immediately calls for a refresh from DB. So it fixes the lag delay bug I got tagged in.

[see here.](https://gph.is/g/4wMv8oj)

Why am I using dangerouslyGetParent? Because we can't access the direct parent Header since we have so many nested navigators. It's not actually dangerous, per documentation, the alternative is to de-stack navigators in AppNavigator. Which I'm not sure is possible if we want to use the Drawer.
["Reason why the function is called dangerouslyGetParent is to warn developers against overusing it to eg. get parent of parent and other hard-to-follow patterns."](https://reactnavigation.org/docs/en/navigation-key.html#usage-with-reset-navigation-actionshtml-reset)


civilapp/src/current/RoomDetails.js
```diff
@@ -134,20 +134,28 @@ class RoomDetails extends React.Component {
...
       this.props.navigation.setParams({
         roomInfo: this.fetchRoomInfo(),
       })
     }
+    
+    if(this.props.navigation.getParam('shouldUpdateHeader')) {
+      refreshScreen()
+      this.props.navigation.setParams({ shouldUpdateHeader: 0 })
+    }
+    if (
```
rauchp: I think we should keep this to guarantee when someone opens the drawer they aren't seeing stale data.

civilapp/src/current/RoomDetails.js
```diff
@@ -134,20 +134,28 @@ class RoomDetails extends React.Component {
...
-
       this.props.navigation.setParams({
         roomInfo: this.fetchRoomInfo(),
       })
     }
+    
+    if(this.props.navigation.getParam('shouldUpdateHeader')) {
+      refreshScreen()
+      this.props.navigation.setParams({ shouldUpdateHeader: 0 })
+    }
```
EphemeralX: It looks like this is checking for existence and doesn't check the value of 1, so this might always update, is that intentional?  Also rather than 0 and 1, it's more verbose to use true and false imo (nit).

rauchp: javascript being javascript, that conditional doesn't activate if that value doesn't exist OR if it's 0. So really it doesn't matter. Will change to true/false tho

PR #127, opened 2019-12-06
various fixes

carteryu:
1. edit details scrn has grayed out save button if username is the same shit
2. game select screen is faster now and no longer loads "live game" text if theres no live games. need to see on TF what performance looks like (i.e. if load speed is acceptable)
3. general cleanup 

re: point 2. theres some comments in gameselectscreen that ill obviously cleanup before pushing. i just need to test when theres live game on tomorrow to make sure shit works be4 it gets in master

re: room details - sorry for linter flex pedro


civilapp/src/current/GameSelectScreen.js
```diff
@@ -291,11 +246,30 @@ class GameSelectScreen extends React.Component {
...
+        ></FlatList>
+      </View>
+    )
+  }
+  render() {
+    return (
+      <View>
+        <ImageBackground source={Background} style={styles.parentContainer}>
+          {this.renderHeader()}
+          <ImageBackground style={styles.container} source={GradientFadeTop} />
```
EphemeralX: Use RetroBackground and SafeAreaView (if you don't get to it in this PR, I'll add it in)

carteryu: Yea will do

PR #136, opened 2019-12-13
SelectBannerScreen full implementation

EphemeralX:
* SelectBannerScreen implementation
* CreateRoomScreen navigates to SelectBannerScreen, which will actually create the room.
* Add bannerImage as an attribute to the Room model.

I don't like using AsyncStorage to pass the bannerImage parameter but it didn't work with a parent setState function.


civilapp/src/browse/SelectBannerScreen.js
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    return (
+      <SuperTextHeader
+        leftButtonPress={backButtonPress}
+        rightButtonPress={this.updateBanner}
+        middleText={'Choose Banner'}
+        leftText={'<<'}
+        rightText={'Next'}
+        disabled={disabled}
```
carteryu: should rename this prop at some point, disabled is a super ambiguous name.

civilapp/src/browse/CreateRoomScreen.js
```diff
@@ -42,44 +42,53 @@ export default class CreateRoomScreen extends React.Component {
...
     let { roomTitle, userInfo } = this.state
+
     const channel = await sbCreateOpenChannel(roomTitle)
     await sbOpenChannelEnter(channel)
 
     if (!userInfo) {
       userInfo = await AsyncStorage.getItem('USER_INFO')
       userInfo = JSON.parse(userInfo)
     }
+    const bannerImage = await AsyncStorage.getItem('BANNER_IMAGE')
```
carteryu: why store in async storage?

civilapp/src/browse/SelectBannerScreen.js
EphemeralX: can enhance later as leftButtonHidden and rightButtonDisabled or something.

civilapp/src/browse/CreateRoomScreen.js
EphemeralX: nav params didn't work, got a better idea?

PR #141, opened 2019-12-16
Android text input fixes

EphemeralX:
* Fix for an issue on Android where the keyboard won't show up on UsernameSetupScreen and CreateRoomScreen. Add touchable space on the text input to focus/show keyboard.
* Fix for the background image shifting up on Android when the keyboard shows up or disappears.
* Fix for missing imports in ProfileDrawer


civilapp/src/browse/CreateRoomScreen.js
```diff
@@ -104,19 +104,30 @@ export default class CreateRoomScreen extends React.Component {
...
+          <TextInput
+            autoCorrect={false}
+            keyboardAppearance="dark"
+            blurOnSubmit={false}
+            value={this.state.roomTitle}
+            onChangeText={text => this.setState({ roomTitle: text })}
+            maxLength={24}
+            selectionColor={'#D74BE0'}
+            autoFocus
+            style={{
```
rauchp: nit: can you put this in a style sheet?

civilapp/src/current/RoomDrawer.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import RoomDetails from './RoomDetails.js'
+import { withNavigation } from 'react-navigation'
+class TestDrawer extends React.Component {
```
rauchp: nit: can you rename class to RoomDrawer?

EphemeralX: whoa, I didn't even touch or see this, wtf

civilapp/src/browse/CreateRoomScreen.js
EphemeralX: sure, calling it hiddenHitbox

PR #145, opened 2019-12-19
small fixes

carteryu:
1. fixed username taken problem
2. cleaned up redundant logic in profiledrawer (also moved it away from some bumblefuck folder)
3. added flair to profile drawer
4. fixed milwaukee bucks logo...


rauchp:
Jesus so many linter changes

PR #149, opened 2019-12-20
NEW(er)! CurrentlyPlayingScreen / RoomView Pt1.

rauchp:
I did a lot of behind the scenes stuff here, so I wanted to push this out and stack.

- Made Sendbird information consistent again, whenever we change info from profileEditScreen.
- Removed RoomView header. New shared header for currentlyPlayingScreen
- Chat looks more like new design, but I'm not done there yet.


civilapp/src/components/chat/MessageItem.js
```diff
@@ -1,48 +1,27 @@
...
-import { View, Text } from 'react-native'
+import { View, Text, Image } from 'react-native'
 import { Icon } from 'react-native-elements'
 import { RetroChatText } from '../SuperText'
 
 
 
-const TextItem = props => {
-  // This might change eventially depending on Josh
-  // So for now keep ternary so I don't forget how it's done.
```
rauchp: All this stuff is stored in the original NEW! RoomView PR and the like, I thought it was just crowding up the file.

civilapp/src/auth/AuthUtils.js
```diff
@@ -7,6 +7,37 @@ import { Auth } from 'aws-amplify'
...
+        console.log('Error updating current user info:', error)
+      }
+    }
+  )
+  updateFlairMetadata(sb, flair)
+}
+
+export const updateFlairMetadata = (sb, flair) => {
+  let user = sb.currentUser
+  let data = { flair }
```
EphemeralX: nit: these can be `const`

civilapp/src/sports/utils/FlairUtil.js
```diff
@@ -249,12 +256,17 @@ const EasternConferenceData = [
...
+const getFlairObject = prefix => {
+
+  if (!prefix) return AnonymousObject
   const flair = FlairDict[prefix]
-  return flair ? flair.link : AnonymousUser
+  return flair ? flair : AnonymousObject
+}
+
+const getFlair = prefix => {
+  if (!prefix) return AnonymousUser
```
EphemeralX: you don't need this check since it already exists inside `getFlairObject` which you're calling in here.

civilapp/src/current/CurrentlyPlayingScreen.js
```diff
@@ -36,7 +35,7 @@ export default class CurrentlyPlayingScreen extends Component {
   componentDidMount = async () => {
     this.requestAudioPermission()
 
-    let userInfo = this.props.navigation.getParam('userInfo')
+    let userInfo = this.props.navigation.getParam('userInfo', {})
```
EphemeralX: any reason we want to return {} as the default? as opposed to undefined or null, jw

civilapp/src/sports/utils/FlairUtil.js
rauchp: I still do, b/c I haven't phased out getFlair yet and they are both used in different contexts, where they might each get passed in ''

PR #150, opened 2019-12-20
Sports tab pt1

carteryu:
TODOs for pt2:

1. blue share drawer
2. dropshadow on the tabs
3. make the 3 dots go underneath the player cards (have no idea how to do this without pos: abs)

also need tomove the sportsdata calls to lambda (will do after this sprint)

feel free to point out UI nits. tested on iphone 8 and 11


civilapp/src/current/CurrentlyPlayingScreen.js
```diff
@@ -15,15 +15,23 @@ import CustomRoomTabBar from './common/CustomRoomTabBar'
...
-
+import { SuperTextHeader } from '../components/SuperText'
 import RoomAPI from '../api/RoomAPI'
 import RoomView from './RoomView'
+import SportsTab from './SportsTab'
 
 //This screen has 2 entrypoints, footer tab: Current->CreateRoomModal (Broadcaster), and Browse Screen's goToBroadcast (Listener), each passing navigation parameters.
 //This class is responsible for managing state of user being broadcaster or listener
+
+// const DarkSuperTextHeader = (props) => {
```
carteryu: think this from pedros branch so will be fixed naturally

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,394 @@
...
+  }
+  componentWillUnmount() {
+    clearInterval(this.interval)
+  }
+  renderPlayersTab = (label, data) => {
+    return (
+      <View
+        tabLabel={label}
+        style={{
+          justifyContent: 'center',
```
rauchp: can you start putting styles in a stylesheet? It's hard reading code otherwise

rauchp:
if this is a visual pr can you post screenshots of what you have so far

carteryu:
yep good point ill upload a vid @rauchp

civilapp/src/current/SportsTab.js
carteryu: dang when working i actually hella hate having to take an extra step to look thru a reference, but yea sure ill do it for this pr

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+          horizontal={false}
+          numColumns={2}
+          data={data}
+          keyExtractor={(_, index) => index + ''}
+          renderItem={({ item, index }) => {
+            return (
+              <PlayerStatCard
+                isLeftSide={index % 2 === 0 ? true : false}
+                playerInfo={item}
+              ></PlayerStatCard>
```
EphemeralX: nit: always use /> when there are no children.  favor less code (same for below BoxScoreRows and anywhere else)

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+          scrollEnabled={false}
+          showsVerticalScrollIndicator={false}
+          horizontal={false}
+          numColumns={2}
+          data={data}
+          keyExtractor={(_, index) => index + ''}
+          renderItem={({ item, index }) => {
+            return (
+              <PlayerStatCard
+                isLeftSide={index % 2 === 0 ? true : false}
```
EphemeralX: do we need this? what are you using this for?

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+          <RetroSportsBallText
+            fontSize={14}
+            shadowSize={2}
+            fontFamily={'Sportsball'}
+            message={'No game selected for this room!'}
+          />
+        </View>
+      )
+    }
+    return (
```
EphemeralX: when the JSX gets too long, split it out into smaller pieces for readability, good opportunity for this is the BoxScoreRows for example

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+          >
+            <View style={{ height: '97.5%', width: '95%' }}>
+              <BoxScoreRow
+                header
+                columnOneText={gameData ? gameData.Status : 'Loading'}
+                columnTwoText="1"
+                columnThreeText="2"
+                columnFourText="3"
+                columnFiveText="4"
+                columnSixText="T"
```
EphemeralX: are these column texts going to be variable? if not we should use more descriptive names

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+    }
+  }
+  tick = async gameId => {
+    if (gameId) {
+      const boxScore = await LiveGameData.fetchGameData(gameId)
+      const gameData = boxScore.Game
+      if (gameData) {
+        this.setState({
+          gameData,
+          //TODO: fix case where gamescore is tied.
```
EphemeralX: this seems like a simple enough change, why not just include it here?

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+          message={props.columnFiveText}
+        ></RetroSportsBallText>
+      </View>
+      <View
+        style={{
+          flex: 0.15,
+          justifyContent: 'flex-end',
+          flexDirection: 'row',
+          alignItems: 'center',
+        }}
```
EphemeralX: you could do a style array with a flex overwrite here. i.e. `style={[styles.rowEntry, { flex: 0.15 }]}`

civilapp/src/current/common/PlayerStatCard.js
```diff
@@ -0,0 +1,213 @@
...
+          justifyContent: 'center',
+          alignItems: 'center',
+        }}
+      >
+        <FlatList
+          scrollEnabled={false}
+          showsVerticalScrollIndicator={false}
+          horizontal={false}
+          numColumns={2}
+          data={['', '', '', '']} // to render 4 stats at a time
```
EphemeralX: should be a better way..

civilapp/src/current/common/PlayerStatCard.js
```diff
@@ -0,0 +1,213 @@
...
+            return <StatBox topText={key} bottomText={val}></StatBox>
+          }}
+        />
+      </View>
+    )
+  }
+  render() {
+    const cardHorizontalMargin = this.props.isLeftSide
+      ? { marginRight: 7 }
+      : { marginLeft: 7 }
```
EphemeralX: there should be a cleaner way to do this without this...

civilapp/src/current/common/PlayerStatCard.js
```diff
@@ -0,0 +1,213 @@
...
+            height: 256,
+            marginTop: 8,
+          },
+          cardHorizontalMargin,
+        ]}
+      >
+        <View style={{ alignItems: 'center', paddingTop: 10 }}>
+          <ImageBackground
+            //TODO: when headshots are in, use props to get this
+            source={SampleHeadshot}
```
EphemeralX: should we just use props.headshot here and pass in SampleHeadshot into props.headshot? seems like it's gotta be done eventually anyways

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+  ActivityIndicator,
+  KeyboardAvoidingView,
+  TouchableOpacity,
+  Modal,
+  ScrollView,
+  ImageBackground,
+  FlatList,
+} from 'react-native'
+
+import { getFlair } from '../sports/utils/FlairUtil'
```
rauchp: asset ordering here

civilapp/src/current/SportsTab.js
```diff
@@ -0,0 +1,372 @@
...
+            shadowSize={2}
+            fontFamily={'Sportsball'}
+            message={'SCORING'}
+          />
+        </View>
+        <View style={styles.centeredView}>
+          <ImageBackground
+            resizeMode="stretch"
+            style={{
+              width: width * 0.89,
```
rauchp: these styles can be in a stylesheet as well

civilapp/src/current/SportsTab.js
rauchp: Do you think you could divide the render into smaller pieces? It's pretty hard working on these files afterwards for the rest of us. See my EditRoomDetails to see how I changed it from a big chunk to smaller sections

civilapp/src/current/SportsTab.js
carteryu: if its on the left side of the flatlist u want right margin and vice versa

civilapp/src/current/SportsTab.js
carteryu: at big companies they intentionally spell it out so its clear where exactly the tag is closed

would prefer not changing it unless you feel rly strongly abt this

civilapp/src/current/SportsTab.js
carteryu: yes? wdym

civilapp/src/current/SportsTab.js
carteryu: true

civilapp/src/current/common/PlayerStatCard.js
carteryu: suggestions? and whats wrong with this if you just want the flatlist to iterate four times?

civilapp/src/current/SportsTab.js
rauchp: AirBnB's linter spazzes out when don't do < />, I prefer a single close w/o children by a lot, not sure how strong Cody feels abs it

civilapp/src/current/SportsTab.js
carteryu: sure

civilapp/src/current/SportsTab.js
carteryu: yep on it

civilapp/src/current/common/PlayerStatCard.js
carteryu: ill just fix it when i get the headshots.

civilapp/src/current/SportsTab.js
carteryu: will do

civilapp/src/current/SportsTab.js
carteryu: this feels so trivial all of us are capable of understanding the code whichever way it is.

i guess after i get codys new linter in itll be fine?

rauchp: I understand what it looks like it just looks ugly af and takes up more space, why have it?

civilapp/src/current/SportsTab.js
carteryu: look at designs it should be clear what these rows refer to

civilapp/src/current/SportsTab.js
carteryu: i already answered why

guess you do feel really strongly abt this. will change now

civilapp/src/current/SportsTab.js
carteryu: its good i just forgot to take out the comment, mb

civilapp/src/current/common/PlayerStatCard.js
carteryu: yea sure, u can do a hidden view with 14 padding in between every two items in the flatllist but i think thats more confusing code. id like the way it is rn more.

other thoughts?

EphemeralX: why can't you just have horizontal margin for each render item? with the width of the parent container wide enough that it looks correct.

civilapp/src/current/SportsTab.js
EphemeralX: name these columns properly in the code.  a dev shouldn't have to look at figma to understand what's meant to be rendered.  looks like it's score per quarter and a total? not sure what final means but you do.  quarter1Score, quarter2Score, quarter3Score, quarter4Score, totalScore?

PR #157, opened 2019-12-23
NEW(er)! RoomView Pt.3

rauchp:
- SpeakerList in RoomView
- Added 'is speaking box'
- Prettier'd my shit up some more. 
- Mic goes green / grey depending on voice channel state
- Misc chat re-sizing (mostly how repeat messages from same user are rendered.)

Tested on iOS and Android emulator. Animation works on android, surprisingly.   
Video of PR: https://drive.google.com/file/d/1Hcz_Z7-0spqgc8Gf8xyAQQTWsuBDLhDt/view

Gonna let Alex do some more nitting before I merge this guy.


civilapp/src/components/WoodenBackground.js
```diff
@@ -1,22 +1,55 @@
 import React from 'react'
-import { ImageBackground, View } from 'react-native'
+import { ImageBackground, View, Text, Platform, UIManager, LayoutAnimation } from 'react-native'
+
 import { SafeAreaView } from 'react-navigation'
 import LightWoodenBackgroundImage from '../../assets/images/light_wood_background.png'
 
-const WoodenBackground = (props) => (
+if (Platform.OS === 'android' && UIManager.setLayoutAnimationEnabledExperimental) {
```
rauchp: I could move this to App.JS, but wherever I put it it's going to look ugly as fuck, thoughts?

civilapp/src/components/WoodenBackground.js
```diff
@@ -1,22 +1,55 @@
...
+  }
+}
 
+const WoodenBackground = props => (
   <React.Fragment>
-    <SafeAreaView forceInset={{ bottom: 'never', top: 'always' }} style={{ flex: 1, backgroundColor: props.color ? props.color : '#1F1F1F' }}>
+    {props.isInVoiceChannel && <SafeAreaView forceInset={{ top: 'always' }} style={{ backgroundColor: '#43B581' }} />}
+    <AlertText expanded={props.isInVoiceChannel} />
+    <SafeAreaView
+      forceInset={{ bottom: 'never', top: props.isInVoiceChannel ? 'never' : 'always' }}
```
EphemeralX: is there a way to simplify this safe area logic? for the isInVoiceChannel case, do we need 2 SafeAreaViews?  maybe we can just set the SafeAreaView prop attributes with the isInVoiceChannel prop and have 1?

civilapp/src/current/RoomView.js
```diff
@@ -458,6 +457,26 @@ class RoomView extends React.Component {
...
   }
 
+  renderSpeakerList = () => (
+    <View style={{ flexDirection: 'column', width: 40, position: 'absolute', right: '1.75%', top: '2.5%' }}>
+      {this.speakerObj.map((speaker, index) => {
+        let isSpeaking = this.isSpeakingDict[speaker.id] ? true : false
+        let isMuted = this.selfMutedDict[speaker.id] || this.othersMutedDict[speaker.id] ? true : false
+        if (speaker.id === this.state.userInfo.id) isMuted = this.isMuted
+        if (isMuted) isSpeaking = false
+        let opacityColor = isMuted ? 0.2 : 0.5
```
EphemeralX: const for these.

also suggest maybe 
`const mutedOpacity = isMuted ? 0.2 : 0.5
`const opacityColor = isSpeaking ? 1 : mutedOpacity`
... `<Image opacity={opacityColor} .../>`

civilapp/src/current/CurrentlyPlayingScreen.js
```diff
@@ -116,30 +109,30 @@ export default class CurrentlyPlayingScreen extends Component {
...
             gameId={room.gameId}
             updateCount={online => (this.onlineCount = online)}
             userInfo={userInfo}
             style={styles.container}
             isOpenChannel={true}
             userType={userType}
             room={room}
-            setModal={(bool) => this.setState({ isInModal: bool})}
+            setInVoiceChannel={bool => this.setState({ isInVoiceChannel: bool })}
+            setModal={bool => this.setState({ isInModal: bool })}
```
EphemeralX: nit: I prefer not to use `bool` as a variable name because in 1. it's not descriptive. 2. a lot of programming languages has bool as a primitive type so I don't think it's good practice to have that possible confusion.

prefer just using `isInModal` and `isInVoiceChannel` here.

civilapp/src/components/WoodenBackground.js
EphemeralX: hm, I think if this is going to persist across other screens beyond the RoomView ones, it should exist in a more general file; app.js could make sense since it is the shared parent of all the screens.   if it's just on CurrentlyPlayingScreen children, maybe it can fit there.  I don't know it's ideal to associate with a background though.

civilapp/src/components/WoodenBackground.js
```diff
@@ -1,22 +1,55 @@
...
 
-const WoodenBackground = (props) => (
+if (Platform.OS === 'android' && UIManager.setLayoutAnimationEnabledExperimental) {
+  UIManager.setLayoutAnimationEnabledExperimental(true)
+}
+
+class AlertText extends React.Component {
+  state = { expanded: false }
+
+  componentDidUpdate(prevProps, _) {
```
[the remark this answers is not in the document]
rauchp: What do you mean toggling onPress? I don't like my solution, but it means I don't have to make a WoodenBackground a component to register the change for an animation toggle. Open to ideas on how else to do it

civilapp/src/components/WoodenBackground.js
rauchp: We already had some loose code in App.JS, so it won't look too bad below that actually

PR #160, opened 2019-12-24
Sports card sharing + other small features

carteryu:
1. sharing card to same or diff room both work afaik (did basic qa on iphone 11. could spend some more time later)
2. room header adjusts based off which tab ur in
3. minor ui improvements

notes:
1. linter flexed on a lot of the files. i commented on my one liner changes so that you dont need to go thru the whole file. ill switch to eslint/prettier after my sportstab tickets r done
2. func handlers / prop drilled player stat data from sportscard -->  currentlyplaying (parent) --> roomview --> chatview. super confusing/messy -wrote some documentation to hopefully help clarify wtf is going on


civilapp/src/App.js
```diff
@@ -24,7 +24,7 @@ import config from './aws-config' //Uses custom config instead of generated aws-
 
 useScreens() //Performance optimization for screens
 Amplify.configure(config)
-Amplify.Logger.LOG_LEVEL = 'DEBUG';
+// Amplify.Logger.LOG_LEVEL = 'DEBUG';
```
carteryu: this is the only thing i changed in appjs

civilapp/src/browse/RoomsView.js
```diff
@@ -32,25 +18,26 @@ class RoomsView extends Component {
...
-      },
-    )
+    this.willFocusSubscription = this.props.navigation.addListener('willFocus', () => {
+      this.getSubscribedRooms()
+    })
     let userInfo = await AsyncStorage.getItem('USER_INFO')
     const gameData = await AsyncStorage.getItem('gameSchedule')
     //TODO: add error check in case key is busted
-    const parsedData = JSON.parse(gameData)
+    if (gameData) {
```
carteryu: this is the only thing i changed in roomsview.

civilapp/src/current/common/CustomTeamTabBar.js
```diff
@@ -39,26 +36,22 @@ const CustomTeamTabBar = createReactClass({
     const fontWeight = isTabActive ? 'bold' : 'normal'
 
     return (
-      <TouchableOpacity
+      <TouchableWithoutFeedback
```
carteryu: maybe ill use <TouchableOpacity> with activeOpacity: 1 if this is the reason this shits broken on android. will investigate later.

civilapp/src/current/SportsTab.js
```diff
@@ -85,23 +87,44 @@ class SportsTab extends React.Component {
...
+    this.toastSettings = {
+      text: 'Shared!',
+      buttonText: 'Ok!',
+      duration: 1500,
+      position: 'bottom',
+      style: { bottom: '100%' }
+    }
+
+    //TODO: bandaid solution for didFocus not triggering upon room creation. need to investigate and clean this up.
+    this.fetchGameInfo()
```
carteryu: makes no sense it literally doesnt trigger if being navigated from the banner select screen. also taking suggestions on better ways to do this?

civilapp/src/current/SportsTab.js
```diff
@@ -299,31 +284,73 @@ class SportsTab extends React.Component {
...
+  setParentShareData = (playerInfo, roomChannel, isOtherChannel) => {
+    if (isOtherChannel) {
+      this.sendCustomComponentToOtherRoom()
+    } else {
+      this.props.changePlayerInfo(JSON.stringify(playerInfo), roomChannel)
+      this.setModalVisible(false, 'none', () => {
+        Toast.show(this.toastSettings)
+      })
+    }
+  }
```
EphemeralX: it's a little unclear from reading this function if the else block is sending the player info to the current channel? from what I can tell, you set state in the parent after stringify (is that necessary?), then toggle modal and then show toast. ... it looks like you're using the state change -> prop drill -> sendCustomComponent in ChatView here.  imo we can make this cleaner by having 1 direct way to send the player card (see comments below), and still use the state update in ChatView for re-rendering.

civilapp/src/current/SportsTab.js
```diff
@@ -299,31 +284,73 @@ class SportsTab extends React.Component {
...
+      await sbOpenChannelEnter(channel)
+      sbSendFileMessage(channel, source, (message, error) => {
+        if (error) {
+          console.log('Error sending getting channel:', error)
+        } else {
+          console.log('Successfully shared:', message)
+          Toast.show(this.toastSettings)
+        }
+      })
+      this.setModalVisible(false)
```
EphemeralX: jw if `this.setModalVisible(false)` should follow the call pattern above where you inserted `Toast.show(...)` as a callback to it, and plug it inside the success branch of `sbSendFileMessage`?

civilapp/src/current/SportsTab.js
```diff
@@ -299,31 +284,73 @@ class SportsTab extends React.Component {
...
+      this.sendCustomComponentToOtherRoom()
+    } else {
+      this.props.changePlayerInfo(JSON.stringify(playerInfo), roomChannel)
+      this.setModalVisible(false, 'none', () => {
+        Toast.show(this.toastSettings)
+      })
+    }
+  }
+
+  sendCustomComponentToOtherRoom = async () => {
```
EphemeralX: to me, this implementation just sends a custom file message to a channel (nit: imo it's better style to put roomChannel in the parameter, makes the function more verbose), so I feel like it should be named `sendPlayerCardToRoom`  but I want to understand the difference between sending it to current room vs other room?

EphemeralX: after reading the ChatView part of this, as noted in a below comment, I think it's better to have this refactored as a utility/helper function in ChatUtils with 3 params: roomChannel, playerInfo, callback.  callback could be your setModalVisible and Toast.  That way, whether you're updating in ChatView or here, you're doing the same thing and reducing code redundancy.

civilapp/src/current/RoomView.js
```diff
@@ -294,6 +275,12 @@ class RoomView extends React.Component {
     this.initializeRoom()
   }
 
+  componentDidUpdate = async prevProps => {
+    if (prevProps.playerInfo && prevProps.playerInfo !== this.state.playerInfo) {
```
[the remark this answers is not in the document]
carteryu: by this pattern u mean componentdidupdate?

yes im listening for prop changes. without comparing the previous prop to the current state it will call itself a lot more times. the check prevents that so only one call is executed

civilapp/src/current/SportsTab.js
```diff
@@ -8,10 +8,13 @@ import {
   ImageBackground,
   FlatList,
   TouchableOpacity,
-  Text,
+  Text
 } from 'react-native'
 
-import { sbGetOpenChannel, sbSendFileMessage } from '../ChatUtils'
+const { TouchableWithoutFeedback } = ReactNative
```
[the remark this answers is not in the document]
carteryu: caught this later and changed locally without pushing

civilapp/src/current/SportsTab.js
carteryu: the else block is sending player info to the current channel. its different logic that needs to update state in chatview directly

carteryu: yes setting state in parent is i think, ideal. setting state triggers a rerender and therefore roomview -> chatview props get updated

civilapp/src/current/SportsTab.js
carteryu: yea i can do that. afaik theyre the same thing but youre right can do it for consistency

civilapp/src/current/SportsTab.js
carteryu: i can update param yea

civilapp/src/current/RoomView.js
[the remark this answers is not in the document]
EphemeralX: why is it ` prevProps.playerInfo !== this.state.playerInfo` instead of ` prevProps.playerInfo !== this.props.playerInfo` though?  should be same # of calls unless I'm missing something?

civilapp/src/current/SportsTab.js
```diff
@@ -85,23 +87,44 @@ class SportsTab extends React.Component {
...
+
+    this.toastSettings = {
+      text: 'Shared!',
+      buttonText: 'Ok!',
+      duration: 1500,
+      position: 'bottom',
+      style: { bottom: '100%' }
+    }
+
+    //TODO: bandaid solution for didFocus not triggering upon room creation. need to investigate and clean this up.
```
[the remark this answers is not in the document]
carteryu: but i need the didfocus on the sports tab not the currentlyplayingscreen. will take a look at the scrollindex thing

PR #161, opened 2019-12-24
Twitter highlights

EphemeralX:
READ 
* Ignore the files that was hit by my auto-linter. (BrowseScreen, CustomRoomTabBar, SportsTab) Sorry about that.
* I tried to make both the HighlightsTab.js and TwitterService/index.js as examples of what I would like to see from you guys in terms of code polish/clean-up, styling, consistency, spacing, nits, etc. (I don't think I write perfect code by any means, and there are some things I'm not happy about in these code samples like the amount of Platform.OS === 'ios' checks I needed, a TODO: note since we will move off client-side sportsdata calls, etc.  Feel free to nit me or let me know if you disagree with what I consider good practice). For additional context on dev code vs polish code, feel free to compare my 4th to last commit vs the previous.

Changes
* Twitter Highlight Feature
* Twitter Lambda update
* ESlinter rules update

Next steps:
* Optimize the call load of TwitterService lambda with REDIS integration (clients will hit REDIS for results)
* Set up scheduler to call the TwitterService lambda to warm the REDIS cache (need to integrate with GameSchedule logic and possibly Sportsdata Lambda)

Tested on iOS 8, X, 11 Sim, Android S8 device

https://streamable.com/l1su6

Android:
https://files.slack.com/files-pri/TMDEZAG4S-FRR2UT1QS/screenshot_20191227-185452.png


civilapp/src/current/HighlightsTab.js
```diff
@@ -0,0 +1,417 @@
...
+              source={{ uri: selectedTweet.videoUrl }}
+              style={styles.video}
+              resizeMode='stretch'
+              paused={!playback}
+            />
+          ) : (
+            <View style={styles.video} />
+          )}
+        </TouchableHighlight>
+        {selectedTweet && (
```
EphemeralX: self nit: can move this code to a `renderVideoCaption()` function

civilapp/src/current/HighlightsTab.js
```diff
@@ -0,0 +1,417 @@
...
+      const awayTeamAlias = getTwitterAlias(awayTeam)
+      const awayTeamTweets = await this.fetchUserTweets(awayTeamAlias, 10)
+
+      const recentTweets = homeTeamTweets.concat(awayTeamTweets)
+      recentTweets.sort((a, b) => {
+        return Date.parse(b.timestamp) - Date.parse(a.timestamp)
+      })
+
+      const videoTweets = recentTweets.filter(tweet => tweet.hasVideo)
+      const regularTweets = recentTweets.filter(tweet => !tweet.hasVideo)
```
EphemeralX: self nit:
```
let videoTweets, regularTweets = []
recentTweets.forEach(tweet => tweet.hasVideo ? videoTweets.push(tweet) : regularTweets.push(tweet))
```

civilapp/src/current/HighlightsTab.js
```diff
@@ -0,0 +1,436 @@
...
+
+import { RetroSportsBallText } from '../components/SuperText'
+
+import { TWEET_API_ENDPOINT } from '../settings'
+import { getTwitterAlias } from '../sports/utils/TeamNameMap'
+import * as LiveGameData from '../sports/LiveGame'
+
+const { width, height } = Dimensions.get('window')
+
+class HighlightsTab extends React.Component {
```
rauchp: Most of the time in our codebase we do
`import React, { Component } from 'react-native'`
`class Class extends Component {`

Do you want us to switch to this?

The docs import Component as well 
https://facebook.github.io/react-native/docs/view

civilapp/src/current/HighlightsTab.js
```diff
@@ -0,0 +1,436 @@
...
+    return (
+      <View style={styles.videoCaption}>
+        <Text selectable style={[styles.captionText, { color: 'white' }]} numberOfLines={1}>
+          {selectedTweet.text}
+        </Text>
+        <View style={{ flexDirection: 'row' }}>
+          <Text selectable style={styles.captionText}>
+            {selectedTweet.user.name} on Twitter{' '}
+          </Text>
+          <Icon name='twitter' size={12} color='#A1A1A1' style={{ padding: 2 }} />
```
rauchp: can't we const this color if we're gonna use it more than once?

civilapp/src/current/HighlightsTab.js
```diff
@@ -0,0 +1,436 @@
...
+        {this.renderVideoContainer()}
+        {this.renderGameHighlights()}
+        <View style={styles.lineSeparator} />
+        {this.renderInternetChatter()}
+      </ScrollView>
+    )
+  }
+}
+
+export default HighlightsTab
```
rauchp: I've always seen export as either at the very end or a file, or as part of class declaration (export default class Class). Pretty uncommon to see it halfway through a file like this.

civilapp/src/current/HighlightsTab.js
```diff
@@ -0,0 +1,436 @@
...
+    top: 2.5,
+    left: 10.5,
+    borderRadius: 14,
+    zIndex: -1,
+    position: 'absolute',
+    height: '99%',
+    width: '97%',
+    backgroundColor: 'black'
+  }
+})
```
rauchp: React Web recommends the convention of having index.js and index.css files, where we use the css files solely to import styles for the .JS sheets to use. Considering 100+ lines of this file are styling, I wonder if we can employ that for larger stylesheets like this.

civilapp/src/current/HighlightsTab.js
```diff
@@ -0,0 +1,436 @@
...
+        </View>
+        <FlatList
+          ref={ref => {
+            this.gameHighlightsFlatList = ref
+          }}
+          contentContainerStyle={{ flexGrow: 1 }}
+          initialNumToRender={4}
+          horizontal
+          showsHorizontalScrollIndicator={false}
+          keyExtractor={(_, index) => `${index}`}
```
rauchp: This keyExtractor is actually a lot neater than what we've been doing 😫, I'm gonna steal it

civilapp/src/current/HighlightsTab.js
EphemeralX: I don't think I'd go that far; I'm not personally a fan of that paradigm - I like having js, jsx, styles in one file.

civilapp/src/current/HighlightsTab.js
EphemeralX: that's true, in this cause I might put it at the top since I'm not using an HoC on it like withNavigation in other places

civilapp/src/current/HighlightsTab.js
EphemeralX: I'm not too picky on this point actually, but I think using Component is cleaner than React.Component.

civilapp/src/current/HighlightsTab.js
EphemeralX: Oh I didn't notice, good point

PR #163, opened 2019-12-28
sports data wrap up

carteryu:
diffing against my old sports data branch.

afk for a bit, doing a writeup soon


civilapp/src/current/GameSelectScreen.js
```diff
@@ -153,32 +143,18 @@ class GameSelectScreen extends React.Component {
...
                           middle={item.AwayTeamScore}
-                          right={displayTime(
-                            item.TimeRemainingMinutes,
-                            item.TimeRemainingSeconds,
-                            item.Status,
-                          )}
-                          rightColor={'red'}
-                        ></TeamStats>
+                          right={displayTime(item.TimeRemainingMinutes, item.TimeRemainingSeconds, item.Status)}
+                          rightColor={'red'}></TeamStats>
```
EphemeralX: nit: I get that you disagree, but disagree and commit.  Once you have the ESLinter/Prettier setup it'll complain on this as well.

civilapp/src/current/SportsTab.js
```diff
@@ -112,23 +132,20 @@ class SportsTab extends React.Component {
...
-      buttonText: 'Ok!',
-      duration: 1500,
-      position: 'bottom',
-      style: { bottom: '100%' }
-    }
+    this.currentRoomId = this.subscribedRooms ? this.subscribedRooms[0].id : ''
+    this.subscribedRooms.forEach(async room => {
+      const onlineCount = await this.getChannelOnlineCount(room.chatChannelUrl)
+      room['onlineCount'] = onlineCount
+    })
```
EphemeralX: just wanna double check that this is tested and doesn't need a Promise.all or anything right?

civilapp/src/current/SportsTab.js
```diff
@@ -165,25 +182,39 @@ class SportsTab extends React.Component {
...
-        renderItem={({ item, index }) => {
-          if (index === 0) this.currentRoomId = item.id
-          return this.renderSubscribedRoomRow(item)
-        }}
-      />
-    )
-  }
+  renderSubscribedRoomList = () => (
+    <FlatList
+      scrollEnabled={true}
```
EphemeralX: nit: prefer just do `scrollEnabled` and it's the same as setting it to true.

civilapp/src/current/SportsTab.js
```diff
@@ -165,25 +182,39 @@ class SportsTab extends React.Component {
...
+  )
+
+  renderTextInputBox = () => (
+    <View>
+      <View style={styles.textInputViewStyle}>
+        <TextInput
+          style={styles.textInputStyle}
+          placeholder={width > 350 ? 'insert big brain commentary here.' : 'insert commentary here.'}
+          placeholderTextColor={'#A1A1A1'}
+          autoCorrect
```
EphemeralX: this might be personal preference but I hate auto correct in chat

civilapp/src/current/SportsTab.js
```diff
@@ -199,6 +230,9 @@ class SportsTab extends React.Component {
   }
 
   renderSubscribedRoomRow = item => {
+    const subText = `${item.onlineCount} ONLINE, ${item.memberCount} ${item.memberCount > 1 ? 'MEMBERS' : 'MEMBER'} ${
```
EphemeralX: nit: I'd factor out the current room logic into it's own variable and append it to subText if relevant - having logic in string interpolation isn't good/clean

civilapp/src/current/SportsTab.js
```diff
@@ -311,20 +368,29 @@ class SportsTab extends React.Component {
   //roomChannel - sendbird channel
   //isOtherChannel - if sending the current channel we need to modify state in chatview, otherwise we need to call it outside chatview (i.e. here)
   setParentShareData = (playerInfo, roomChannel, isOtherChannel) => {
```
EphemeralX: we should rename this to something more self-documenting... sendPlayerInfoCard?

civilapp/src/current/SportsTab.js
```diff
@@ -353,13 +433,28 @@ class SportsTab extends React.Component {
     this.setState({ modalVisible: bool, selectedPlayerInfo: playerInfo }, callback)
   }
 
+  onRefresh = () => {
+    this.setState({ refreshing: true })
+    this.fetchGameInfo()
+    this.setState({ refreshing: false })
```
EphemeralX: setState is asynchronous so you want to do something like:
```
this.setState({ refreshing: true}, () => this.fetchGameInfo())
```
and then set refreshing to false at the end of `this.fetchGameInfo()`

civilapp/src/current/SportsTab.js
```diff
@@ -521,49 +616,93 @@ const BoxScoreRow = props => {
...
-        </ImageBackground>
-      </TouchableOpacity>
-    </View>
+    <TouchableOpacity
+      onPress={() => props.onPress()}
+      disabled={props.disabled}
+      style={{
+        alignItems: 'center'
+      }}>
+      <ImageBackground resizeMode='stretch' source={BlueButton} style={styles.submitButton}>
```
EphemeralX: nit: TODO: We should code up the blue button rather than rely on the asset.

civilapp/src/current/SportsTab.js
```diff
@@ -521,49 +616,93 @@ const BoxScoreRow = props => {
...
+          mainColor={props.disabled ? 'gray' : '#00FAFF'}
+          message={'FIRE AWAY'}
+        />
+      </ImageBackground>
+    </TouchableOpacity>
   )
 }
 
+const isNullOrEmpty = str => {
+  return !str || !str.trim()
```
EphemeralX: is `!str.trim()` equivalen to `str !== ''` ?

civilapp/src/current/SportsTab.js
```diff
@@ -521,49 +616,93 @@ const BoxScoreRow = props => {
...
+          style={{
+            borderRadius: 14,
+            borderWidth: 1.5,
+            borderColor: 'black',
+            backgroundColor: '#E9E9EB',
+            height: 115,
+            width: 270,
+            flexDirection: 'column',
+            paddingVertical: 7,
+            paddingHorizontal: 12
```
EphemeralX: nit: long -> stylesheet

civilapp/src/current/SportsTab.js
```diff
@@ -266,42 +314,51 @@ class SportsTab extends React.Component {
...
-          />
-        </RetroModal>
-      </React.Fragment>
+            style={{
+              height: 80,
+              width: '100%',
+              justifyContent: 'center',
+              alignItems: 'center',
+              backgroundColor: '#0D0B5B',
+              paddingBottom: 10
```
EphemeralX: nit: long -> stylesheet

civilapp/src/components/RetroModal.js
```diff
@@ -29,40 +37,52 @@ const RetroModalFooter = ({ props }) => (
...
+          <RetroModalHeader props={props} />
+          <ImageBackground
+            source={Background}
+            style={{
+              width: '100%',
+              flexGrow: 1,
+              height: props.childrenHeight,
+              flexDirection: 'column',
+              justifyContent: 'flex-start',
+              alignItems: 'center'
```
EphemeralX: nit: long -> stylesheet

civilapp/src/components/RetroModal.js
```diff
@@ -29,40 +37,52 @@ const RetroModalFooter = ({ props }) => (
...
           style={{
-            width: '100%',
+            borderTopLeftRadius: 14,
+            borderTopRightRadius: 14,
+            backgroundColor: props.backgroundColor ? props.backgroundColor : '#1F1F1F',
             flexGrow: 1,
+            width: '100%',
             flexDirection: 'column',
             justifyContent: 'flex-start',
             alignItems: 'center'
```
EphemeralX: nit: long -> stylesheet

civilapp/src/social/ChatView.js
```diff
@@ -83,10 +83,11 @@ class ChatView extends Component {
...
     //Prop drilled all the way from SportsTab. Used to render new sports image
-    if (prevProps.playerInfo && prevProps.playerInfo !== this.state.playerInfo) {
-      this.setState({ playerInfo: prevProps.playerInfo, selectedRoom: prevProps.selectedRoom }, () => {
-        this.sendCustomComponent(this.state.playerInfo, this.state.selectedRoom)
-      })
+    if (this.props.playerInfo !== prevProps.playerInfo) {
+      setTimeout(() => {
+        this.initChat()
+        this.setState({ newComponentPassed: true })
+      }, 1000)
```
EphemeralX: what is the purpose of setTimeout here?

civilapp/src/social/ChatView.js
```diff
@@ -315,8 +316,13 @@ class ChatView extends Component {
         const messages = await sbGetMessageList(previousMessageListQuery)
         this.setState({
           previousMessageListQuery,
-          messageList: this.uniqueList([...this.state.messageList, ...messages])
+          messageList: this.state.newComponentPassed
+            ? this.uniqueList([...messages, ...this.state.messageList])
+            : this.uniqueList([...this.state.messageList, ...messages])
```
EphemeralX: why do you need to reverse the order of messageList and messages here?

civilapp/src/social/ChatView.js
```diff
@@ -83,10 +83,11 @@ class ChatView extends Component {
     }
 
     //Prop drilled all the way from SportsTab. Used to render new sports image
-    if (prevProps.playerInfo && prevProps.playerInfo !== this.state.playerInfo) {
-      this.setState({ playerInfo: prevProps.playerInfo, selectedRoom: prevProps.selectedRoom }, () => {
-        this.sendCustomComponent(this.state.playerInfo, this.state.selectedRoom)
-      })
+    if (this.props.playerInfo !== prevProps.playerInfo) {
```
EphemeralX: my understanding is that prop changes -> setState -> re-render so that makes sense.  but i don't understand the other parts of this method, are you sure the other parts are necessary? (setTimeout, switching ...messages and ...this.state.messageList)?

civilapp/src/current/GameSelectScreen.js
carteryu: yea sg

civilapp/src/current/SportsTab.js
carteryu: yea works well to my knowledge. why do you say this might need promise.all?

civilapp/src/current/SportsTab.js
carteryu: kk

civilapp/src/current/SportsTab.js
carteryu: Keeping it consistent with how pedro did his textinput. if i take it out i should take it out from his as well

civilapp/src/current/SportsTab.js
carteryu: yea

civilapp/src/current/SportsTab.js
carteryu: yep i know. will be in next pr

civilapp/src/current/SportsTab.js
carteryu: oh, good catch. althought this does seem to work fine, ill still change it since i agree with what ur saying.

civilapp/src/current/SportsTab.js
carteryu: on it

civilapp/src/current/SportsTab.js
carteryu: not a nit, i should be doing this and its on my notion

civilapp/src/current/SportsTab.js
EphemeralX: JS is weird so I'm just paranoid that not all rooms are updated by the time you process this.subscribedRooms later.

civilapp/src/current/SportsTab.js
carteryu: wait so where do you suggest i build this string? i need access to the item object

civilapp/src/components/RetroModal.js
```diff
@@ -29,40 +37,52 @@ const RetroModalFooter = ({ props }) => (
 
 const RetroModal = props => (
   <Modal transparent animationType='slide' visible={props.isVisible}>
-    <View
+    <KeyboardAvoidingView
+      enabled={props.hasKeyboard}
+      behavior={Platform.OS === 'ios' ? 'position' : undefined}
```
rauchp: in another file I made the platform check into a const up top.

so const isIOS = platfoorm.OS ===
and then the start of the code

civilapp/src/current/GameSelectScreen.js
```diff
@@ -153,32 +143,18 @@ class GameSelectScreen extends React.Component {
...
                         <TeamStats
                           left={item.HomeTeam}
                           middle={item.HomeTeamScore}
                           right={QuarterMap[item.Quarter]}
-                          rightColor={'#8086A9'}
-                        ></TeamStats>
+                          rightColor={'#8086A9'}></TeamStats>
```
rauchp: Didn’t we all agree to do < /> for components without children

civilapp/src/current/GameSelectScreen.js
```diff
@@ -188,8 +164,7 @@ class GameSelectScreen extends React.Component {
                 </TouchableOpacity>
               )
             }
-          }}
-        ></FlatList>
+          }}></FlatList>
```
rauchp: Same here with the < />

civilapp/src/current/SportsTab.js
EphemeralX: right above subText.

civilapp/src/current/GameSelectScreen.js
carteryu: yea cody nitted on me already i fixed it

civilapp/src/components/RetroModal.js
carteryu: oh ok doing it now

carteryu:
> 2 things:
> 
> I see you changed the retro modal, did you test to make sure it still works as the voice channel modal? Something already broke the chatView and now entering the chat pushes the input out of roomview out of bounds, don’t want anymore breaking changes to an important part of the app.
> 
> Can you put a bit more detail in the description of the PR next time?
> 
> Otherwise, looks good!

sry i came back too late last night and i saw cody already left his nits so i didnt end up doing the writeup.mb

PR #167, opened 2019-12-29
Polish Currently twitter

rauchp:
- Fixed overlayBottom so dots display over content.
- Changed how we do flatlist for twitter page.

2_pr

2_pr

2_pr

2_pr


civilapp/src/current/CurrentlyPlayingScreen.js
```diff
@@ -126,8 +126,8 @@ export default class CurrentlyPlayingScreen extends Component {
         <ScrollableTabView
           initialPage={1}
           prerenderingSiblingsNumber={3}
-          tabBarPosition='bottom'
-          renderTabBar={() => <CustomRoomTabBar />}
+          tabBarPosition='overlayBottom'
```
carteryu: nice, quesiton - do you need this since your passing overlaybottom as prop and doing a whole bunch of shit with it?

rauchp: tabBarPosition is what actually absolutely overlays it, but it did some weird shit where the width wasn't being set, so I had to add the props afterwards to style it when we do overlayBottom

EphemeralX: the prop below is for the styles in CustomRoomTabBar, but this setting is needed on the ScrollableTabView level

carteryu: i see now, dope!

civilapp/src/current/HighlightsTab.js
```diff
@@ -199,9 +200,15 @@ export default class HighlightsTab extends Component {
   renderGameHighlights() {
     const { videoTweets } = this.state
     return (
-      <View style={{ flex: 0.215 }}>
-        <View style={{ padding: '4%', paddingBottom: '2%' }}>
-          <SuperTextWithShadow fontSize={10} shadowSize={2} fontFamily='Sportsball' message='Game Highlights' />
+      <View>
+        <View style={{ alignSelf: 'flex-start', padding: '4%', paddingBottom: '2%' }}>
```
EphemeralX: FYI, locally I made this padding 3% and paddingBottom 1% to look closer to the figma designs.  optional for you to add it in this PR, since when I make my PR it'll show.

EphemeralX: also did alignSelf: 'flex-start' work on the view? I had it in SuperTextWithShadow's containerStyle but this is also fine if it works

civilapp/src/current/HighlightsTab.js
```diff
@@ -304,21 +308,28 @@ export default class HighlightsTab extends Component {
     )
   }
 
+  renderListHeader = () => (
+    <React.Fragment>
+      {this.renderVideoContainer()}
+      {this.renderGameHighlights()}
+      <View style={styles.lineSeparator} />
+      <View style={{ paddingLeft: '4%', alignSelf: 'flex-start' }}>
```
EphemeralX: 3% to match my above comment

civilapp/src/current/HighlightsTab.js
```diff
@@ -304,21 +308,28 @@ export default class HighlightsTab extends Component {
...
-    return (
-      <ScrollView
-        contentContainerStyle={{ flex: 1 }}
-        refreshControl={<RefreshControl refreshing={refreshing} onRefresh={this.onRefresh} />}>
-        {this.renderVideoContainer()}
-        {this.renderGameHighlights()}
-        <View style={styles.lineSeparator} />
-        {this.renderInternetChatter()}
-      </ScrollView>
-    )
```
EphemeralX: does the scroll to refresh work the same on the flatlist if we remove this?

EphemeralX:
btw small UI nit: can we make the 3 dots have symmetrical padding top and bottom from the apple drawer thing? it looks unbalanced

civilapp/src/current/HighlightsTab.js
rauchp: Yep worked and it seemed simpler, although we could remove the outer view and just containerStyle as well.

civilapp/src/current/HighlightsTab.js
rauchp: Yep, just added refreshControl to the flatlist

PR #173, opened 2019-12-30
Polish PR

rauchp:
Going to do a number of small fixes and put them all in one PR.
 - Fixed Touchable Opacity for cards 

2_pr

- Fixed Image shifting on transition.

  - https://gph.is/g/4Lgong3

- Fixed Deaf Mode

- Simplified variable names in RoomView to something less confusing 
- Made CreateRoomScreen load faster.
- Fixed Back Button in CurrentlyPlayingScreen after making a room (now takes you back to Browse Screen rather than Select Banner.)
- Fixed ChatView loading.


civilapp/src/browse/common/SportsCardNew.js
```diff
@@ -92,8 +92,9 @@ class SportsCardNew extends Component {
             {this.renderTitle()}
             {this.renderSubtitle()}
           </ImageBackground>
+          <View style={[styles.dropShadow, widthStyle]} />
         </TouchableOpacity>
-        <View style={[styles.dropShadow, widthStyle]} />
+
```
EphemeralX: damn, why didn't I try this...

civilapp/src/landing/LoadingTransition.js
```diff
@@ -18,6 +17,14 @@ const loadingText = require('../../assets/images/loading_text.png')
...
 const { width, height } = Dimensions.get('window')
 
+const InsetBackground = props => {
+  const insets = useSafeArea()
+  return (
+    <ImageBackground imageStyle={{ marginTop: insets.top }} source={RushBackground} style={styles.center}>
+      {props.children}
+    </ImageBackground>
+  )
+}
```
EphemeralX: nice find - didn't even think to look here...

civilapp/src/current/RoomView.js
```diff
@@ -160,9 +154,7 @@ class RoomView extends React.Component {
     if (onlineUsers) {
       let speakers = room.speakers ? room.speakers : []
       speakers = speakers.filter(async speakerId => {
-        const speakerInRoom = onlineUsers.find(user => {
-          user.userId === speakerId
-        })
+        const speakerInRoom = onlineUsers.find(user => user.userId === speakerId)
```
rauchp: this was actually breaking a lot

rauchp:
In CreateRoomScreen, by taking out the setState in ComponentDidMount, it sped up the screen by half a second on loading. In general I’ve been trying to avoid setStates as much as possible since I’m a big app like this where we have several sub nodes in the dom (thanks to SuperText, Inputs, and other advanced components) and setState on these sub nodes is really expensive. 

For the chatView, it was initialized to never start with isLoading, so it just wasn’t working.

civilapp/src/current/RoomView.js
rauchp: this one line fixed like 2 bugs

PR #174, opened 2020-01-03
RIP blue button

carteryu:
Took out blue_button.png and replaced with code

Tested every screen touched on iphone 8 and 11

Only thing thats busted is this screen on iphone 8 b/c i took out justifyContent: 'flex-end' which messes up my gradientbutton component and i think the whole screen should be reworked anyway cuz on small screens the proportions are wayy off regardless of my change
 
Screen Shot 2020-01-02 at 5 55 34 PM


civilapp/src/components/GradientButton.js
```diff
@@ -4,18 +4,38 @@ import { SuperTextWithShadow } from './SuperText'
...
+          <SuperTextWithShadow
+            message={props.text ? props.text : ''}
+            mainColor={props.textColor ? props.textColor : '#FFFFFF'}
+            fontSize={props.fontSize ? props.fontSize : 10}
+            shadowSize={props.shadowSize ? props.shadowSize : 1}
+            fontFamily={'Sportsball'}
+            lightDropShadow={props.lightDropShadow}
+            mediumDropShadow={props.mediumDropShadow}
+          />
+        )}
```
EphemeralX: nit: in this case, I think doing { props.children || (<SuperTextWithShadow .../>) } is more elegant.  The only time I wouldn't do this is if the parent container only accepts 1 child, but this is a View so should be fine.

civilapp/src/auth/ConfirmSignInScreen.js
```diff
@@ -137,13 +136,17 @@ export default class ConfirmSignInScreen extends Component {
...
-          <Text style={styles.resendButtonText}>I didn't get the dumb code</Text>
-        </ImageBackground>
-      </TouchableOpacity>
+      <GradientButton
+        height={39}
+        width={width * 0.8}
+        leftShadowOffset={4}
+        topShadowOffset={4}
+        borderRadius={8}
+        onPress={() => this.resendCode()}
```
EphemeralX: wouldn't it be too much trouble to refactor this into a function outside and call it here while you're touching these files?
re: https://medium.com/@adamjacobb/react-native-performance-arrow-functions-binding-3f09cbd57545

civilapp/src/current/GameSelectScreen.js
```diff
@@ -104,11 +101,13 @@ class GameSelectScreen extends React.Component {
             />
           </View>
         </View>
-        {this.state.doneLoading ? null : (
-          <View style={{ width: width }}>
-            <ActivityIndicator size='large' color='#FFFFFF' style={{ width: '100%', paddingTop: '5%' }} />
-          </View>
-        )}
+        <>
+          {this.state.doneLoading ? null : (
```
EphemeralX: let's get into the habit of deconstructing state everywhere please.

civilapp/src/current/GameSelectScreen.js
```diff
@@ -120,72 +119,59 @@ class GameSelectScreen extends React.Component {
...
             if (item.Status === cardStatus) {
               return (
-                <TouchableOpacity
-                  style={{
-                    width: width * 0.9,
-                    height: HeightOfCard,
-                    marginHorizontal: width * 0.05,
-                    marginVertical: 7.5
+                <GradientButton
+                  onPress={() => this.selectGame(item.GameID)}
```
EphemeralX: factor out please?

civilapp/src/components/GradientButton.js
carteryu: oh shit if this doesnt break anything ill do that.

civilapp/src/auth/ConfirmSignInScreen.js
carteryu: kk u right, sg

civilapp/src/current/GameSelectScreen.js
carteryu: kk

civilapp/src/components/GradientButton.js
```diff
@@ -4,18 +4,38 @@ import { SuperTextWithShadow } from './SuperText'
 export default GradientButton = props => {
   //The default blue used everywhere in figma
   const backgroundColor = props.backgroundColor ? props.backgroundColor : '#020E53'
+  const topShadowOffset = props.topShadowOffset ? props.topShadowOffset : 5
+  const leftShadowOffset = props.leftShadowOffset ? props.leftShadowOffset : 5
+  //For some reason, figma uses BR 10 for everything except for roomview which is 14
```
rauchp: can you remove this comment? doesn't seem relative in the long run

carteryu: i thought itd be helpful if someone uses the component

carteryu: but yea sure

PR #177, opened 2020-01-03
Refactor and slight App performance stuff

EphemeralX:
General changes include switching to PureComponents, moving arrow function definitions out of render functions, removing state variables where possible, cleaning up how we get userInfo, general refactor/linter stuff.

* App, AppNavigator, AuthLoading, AuthUtils, Browse, Discover, Landing, LoadingTransition touched so far


lambdas/updateSubscribedRoomsResolver/index.js
```diff
@@ -0,0 +1,43 @@
+const AWS = require('aws-sdk')
```
carteryu: does this mean you figured out how to replace shitty vtls?

civilapp/src/browse/RoomsView.js
```diff
@@ -79,9 +77,12 @@ class RoomsView extends Component {
       // Sort by online count
       const roomsWithOnlineCount = await Promise.all(
         subscribedRooms.map(async room => {
-          const query = await sbCreateParticipantListQuery(room.chatChannelUrl)
-          const onlineUsers = await sbGetParticipantList(query)
-          room.onlineCount = onlineUsers.length
+          console.log('Room:', room)
```
carteryu: is this necessary?

civilapp/src/current/RoomView.js
```diff
@@ -68,7 +69,13 @@ class RoomView extends React.Component {
 
     if (this.isInVoiceChannel) {
       AgoraAPI.stopBroadcasting()
-      RoomAPI.removeSpeaker(room.id, userInfo.id)
+      RoomAPI.getRoom(room.id)
```
rauchp: why?

lambdas/updateSubscribedRoomsResolver/index.js
EphemeralX: yes

civilapp/src/browse/RoomsView.js
EphemeralX: oh whoops, no tanymore

civilapp/src/current/RoomView.js
EphemeralX: found a bug where if I delete a room, it keeps an entry in the DB with id and speaker -1 because of this call.  so I have to check if the room still exists before calling removeSpeaker so there is no entry in the DB when a room gets deleted.

lambdas/updateSubscribedRoomsResolver/index.js
EphemeralX: and I bled for this. xD

PR #178, opened 2020-01-03
Misc. Polish PR V2

rauchp:
- Fixed VoiceChannelModal on Android
- Amplify Codegen'd
- Added parameters to subs in GraphQL Schema.
- Users not in RoomView but in speaker list no longer appear. (Need to further optimize in next PR.)
- Fixed preloaded Room loading
  - https://drive.google.com/file/d/1drABxTPSDIVH2QVzEZYp74tEskOIAVGI/view?usp=sharing
  - Had to fix how we transferred props (navigation was passing to grandparent so you can't getParam) , and how we updated them (async meant a promise was being returned.)
  - Changed how we received props (getDerivedStateFromProps), 1 of the reasons it wasn't working before was because prop wouldn't load in time.
  - Why I used getDerivedStateFromProps, https://stackoverflow.com/a/49451707/2976676. tl;dr performance reasons and one less setState 

- Fixed FlairSelectScreen, made it more like Figma
  
  
  
Previously: 
2_pr

Now:
2_pr


civilapp/amplify/backend/api/civilapp/schema.graphql
```diff
@@ -79,8 +79,8 @@ type Mutation {
...
 
 type Subscription {
-  onJoinRoom: Room @aws_subscribe(mutations: ["joinRoom"])
-  onLeaveRoom: Room @aws_subscribe(mutations: ["leaveRoom"])
-  onAddSpeaker: Room @aws_subscribe(mutations: ["addSpeaker"])
-  onRemoveSpeaker: Room @aws_subscribe(mutations: ["removeSpeaker"])
+  onJoinRoom(id: $id): Room @aws_subscribe(mutations: ["joinRoom"])
+  onLeaveRoom(id: $id): Room @aws_subscribe(mutations: ["leaveRoom"])
+  onAddSpeaker(id: $id): Room @aws_subscribe(mutations: ["addSpeaker"])
+  onRemoveSpeaker(id: $id): Room @aws_subscribe(mutations: ["removeSpeaker"])
```
EphemeralX: I don't think `$id` works here, you probably need `String` or `ID`

EphemeralX: make sure you try it before you push though

rauchp: ah my bad, wasn't paying attention while I was copying and pasting from other func

civilapp/src/browse/BrowseScreen.js
```diff
@@ -26,18 +26,27 @@ class BrowseScreen extends Component {
       loading: false,
       userInfo: {}
     }
+
+    const parentNav = props.navigation.dangerouslyGetParent()
+    if (parentNav) {
+      const grandParentNav = parentNav.dangerouslyGetParent()
+      if (grandParentNav) {
+        const preloadedPromise = grandParentNav.getParam('preloadedRooms', [])
```
carteryu: hey j/w any reason to use promise over async await here?

rauchp: Async Await does like the name implies and holds up a thread. I don't want to hold up the constructor (pretty sure it's an anti pattern?) but I do want that promise processed ASAP. I instruct the component to resolve the promise when it's done at comp initialization (before mounting too!) and let the rest of the component lifecycle take place

EphemeralX: I actually fixed some of this in my local branch cause I hastily introduced this originally, but I didn't do the grandparent thing. In LoadingTransition I made it await so that the promise is fulfilled during the 4 second loading clamp and it just needs to be passed thru here successfully.  I also made it a class variable instead of a state. Combining that and this should make this cleaner.

civilapp/src/browse/RoomsView.js
```diff
@@ -127,9 +134,14 @@ class RoomsView extends Component {
 
   renderRoom = room => {
     const { parsedData } = this.state
-    const dataRes = parsedData.filter(entry => {
-      return entry.GameID === room.gameId
-    })
+    let dataRes
+
+    if(parsedData) {
+      let gameId = room.gameId ? room.gameId : null
```
EphemeralX: not sure if we need the `? room.gameId : null` part

civilapp/src/components/RetroModal.js
```diff
@@ -49,10 +49,13 @@ const RetroModal = props => (
...
-        onPress={() => props.toggleModal()}
-        style={{ flex: 0.44, width: '100%' }}
-      />
+      {isIOS ? (
+        <View style={{ flex: 0.44 }}>
+          <TouchableOpacity onPress={() => props.toggleModal()} style={{ height: '100%', width: '100%' }} />
+        </View>
+      ) : (
+        <TouchableOpacity onPress={() => props.toggleModal()} style={{ flex: 0.44, width: '100%' }} />
+      )}
```
EphemeralX: any  reason that we can't just use TouchableOpacity by itself on both platforms?

rauchp: on iOS the touchable without outer view doesn't stretch / grow, and the opposite problem happens on Android with View and Touchable. This is because React Native hacked together the android TouchableOpacity implementation, that's why stuff like [TouchableNativeFeedback](https://facebook.github.io/react-native/docs/touchablenativefeedback) exists. The hackiness means they didn't get the platforms to act the same way for the same components when it comes to touchables, whoops.

PR #179, opened 2020-01-04
activity spinners + refactoring

carteryu:
1. reorg'd roomdetails and sportstab
2. when saving to DB, youll now see activity spinners. video avail upon request but its pretty easy to see and basically what ig does
3. added hitslop to headers + one of my buttons that felt hard to hit while i was testing on TF (https://facebook.github.io/react-native/docs/touchablewithoutfeedback#hitslop)
4. fixed the id null warning thing in roomview (lmk if you still get it after the pr pedro)


civilapp/src/current/CurrentlyPlayingScreen.js
```diff
@@ -111,10 +111,8 @@ export default class CurrentlyPlayingScreen extends Component {
         rightButtonStyle={{ minWidth: 50 }}
         middleButtonStyle={{ maxWidth: width - 100 }}
         leftButtonPress={() => {
-          if(navigation.getParam('fromSelectBanner', false))
```
carteryu: ignore dis, accidentally commited linter flex

civilapp/src/components/SuperText.js
```diff
@@ -423,7 +423,7 @@ const SuperTextHeader = props => (
   <View style={[styles.headerViewStyle, props.headerViewStyle]}>
     {/* Back Button */}
     <View style={[styles.leftButtonStyle, props.leftButtonStyle]}>
-      <TouchableOpacity onPress={props.leftButtonPress}>
+      <TouchableOpacity hitSlop={{ top: 20, left: 20, bottom: 20, right: 20 }} onPress={props.leftButtonPress}>
```
rauchp: holy shit I've always used padding to increase the hitbox but hitSlop sounds funny and works better lmao

civilapp/src/current/SportsTab.js
```diff
@@ -219,9 +204,9 @@ class SportsTab extends React.Component {
   }
 
   renderSubscribedRoomRow = item => {
-    const subText = `${item.onlineCount} ONLINE, ${item.memberCount} ${item.memberCount > 1 ? 'MEMBERS' : 'MEMBER'} ${
-      this.props.roomId === item.id ? '(CURRENT ROOM)' : ''
-    }`
+    const subText = `${item.onlineCount} ONLINE, ${item.members.length} ${
+      item.members.length > 1 ? 'MEMBERS' : 'MEMBER'
+    } ${this.props.roomId === item.id ? '(CURRENT ROOM)' : ''}`
```
EphemeralX:
```
const membersText = item.members.length > 1 ? 'MEMBERS' : 'MEMBER'
const currentRoomText = this.props.roomId === item.id ? '(CURRENT ROOM)' : ''
const subText = `${item.onlineCount} ONLINE, ${item.memberCount} ${membersText} ${currentRoomText}`
```
This is what I wanted you to do when I called this out in a previous PR. ^ . Point is to put logic outside of string interpolation and try to keep the string interpolation as simple as possible.

civilapp/src/current/SportsTab.js
```diff
@@ -113,9 +112,11 @@ class SportsTab extends React.Component {
   }
 
   fetchGameInfo = async () => {
+    console.log('component mounted: fetching game info')
```
rauchp: 👀 console log

civilapp/src/current/RoomDetails.js
```diff
@@ -403,173 +405,182 @@ class RoomDetails extends React.Component {
...
+                        avatar={item.profileImageUrl}
+                        textColor={'#FFFFFF'}
+                        style={{
+                          marginLeft: LeftwardMargin,
+                          marginVertical: 4.5
+                        }}
+                      />
+                    )
+                  }
+                }}
```
rauchp: if this were before January 2nd (black thursday), I would've told u that u can put this into an arrow function and class it as 

renderUserRow = () => (
)

and then renderItem=(({item}) => this.renderUserRow(item)

but I'm still not 100% sure if that reduces performance, so I'll look into it rn and reply to this if you should do that

civilapp/src/components/SuperText.js
```diff
@@ -471,19 +470,27 @@ const SuperTextHeader = props => (
...
-        message={props.rightText}
-        fontSize={12}
-        shadowSize={2}
-        mediumDropShadow
-        fontFamily='Sportsball'
-        mainColor={props.disabled ? '#8086A9' : 'white'}
-      />
-    </TouchableOpacity>
+    {props.saving ? (
+      <>
```
rauchp: what is this <> stuff

civilapp/src/profile/ProfileEditScreen.js
```diff
@@ -104,29 +107,30 @@ class ProfileEditScreen extends React.Component {
...
-    }
-    updateSendBirdInfo(username, profileImageUrl, flair)
-    await UserAPI.updateUser({
-      id: userId,
-      username,
-      flair,
-      profileImageUrl
+      this.setState({ saving: false })
+      const { navigation } = this.props
+      navigation.pop()
```
EphemeralX: not sure if this is necessary, but in case there is setState on unmounted component warning, maybe we should wrap this in the function parameter of setState here as well.

civilapp/src/current/SportsTab.js
carteryu: oopz

civilapp/src/components/SuperText.js
EphemeralX: React.Fragment short-hand.  Learned it from ESLinter.

civilapp/src/components/SuperText.js
carteryu: LOLya

civilapp/src/current/RoomDetails.js
EphemeralX: I've just been doing normal functions for functions that render and calling it as regular functions - I believe that's what you're supposed to do.  Arrow functions for everything else though.

carteryu: how are you measuring performance?

happy to change it if it actuall ydoes

civilapp/src/components/SuperText.js
EphemeralX: I didn't understand top, left, bottom, right until recently lol.

civilapp/src/components/SuperText.js
carteryu: fragments, learned it from ur pr owo..

civilapp/src/current/RoomDetails.js
rauchp: ok from what I've read in those 4 minutes it seems the secret sauce is to

renderItem={this.renderUserRow}

and then

renderUserRow = () => ()

because the arrow isn't inside the rendered object (and by extension, the render function), it's performant and not a problem.

civilapp/src/current/GameSelectScreen.js
```diff
@@ -47,8 +48,11 @@ class GameSelectScreen extends React.Component {
     })
   }
   updateGameId = async (id, gameId) => {
-    await RoomAPI.updateRoom({ id, gameId })
-    this.props.navigation.goBack()
+    this.setState({ saving: true }, async () => {
+      await RoomAPI.updateRoom({ id, gameId })
+      this.props.navigation.goBack()
+      this.setState({ saving: false })
```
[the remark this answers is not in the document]
carteryu: oh interesting. i have never gotten a warning. and im ok with goback and setting saving to false at the same time (miniscule speed performance)

civilapp/src/current/RoomDetails.js
```diff
@@ -403,173 +405,182 @@ class RoomDetails extends React.Component {
...
+          showsVerticalScrollIndicator={false}
+          contentContainerStyle={{
+            paddingBottom: 20,
+            alignItems: 'center',
+            width
+          }}>
+          <GradientButton
+            onPress={() => {
+              this.props.navigation.navigate('GameSelect')
+            }}
```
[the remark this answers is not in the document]
carteryu: dam i got boomed+

civilapp/src/current/RoomDetails.js
carteryu: link me sauce?

PR #180, opened 2020-01-04
RoomView Polish

rauchp:
- Fixed Scroll Offset in chat, sending multiple messages now correctly offsets to the very last message, not 2nd to last.
- Redid MesssageInput to make it f a s t, and now the send button doesn't extend out of the input box when typing a long message. Also looks slightly more like Figma since I removed the drop shadows.
- Optimized RoomView loading and used promises to stop bottlenecking of the view load during mount. Now it's f a s t
- Added a better leave room flow. Now we unsubscribe from sub'd actions and leave the correct way from Speaker List and Agora.
- Added animations to RoomView loading.
   - https://drive.google.com/file/d/1Y5iPY-opjvBlD1JusujUFq1e07SfNfrj/view?usp=sharing
   - Alex fine w/ it, we might need to fine-tune the speed and exactly opacity of it all later, but it's a one-liner.
- Changed hitSlop after testing and feeling the chat on-device.
- Fixed offset for KeyboardAvoidingView when the green in-channel alert is on screen.

Known issues: Joining a room as speaker is S L O W now (but accurate.) Need to optimize SpeakerList or see what I can do with Agora about my speaker list problem.


civilapp/src/components/chat/MessageInput.js
```diff
@@ -11,23 +11,34 @@ import SendButton from '../../../assets/images/chat/send_btn.png'
...
+    this.renderMic = this.renderMic.bind(this)
+    this.renderInputButton = this.renderInputButton.bind(this)
+    this.renderTextInput = this.renderTextInput.bind(this)
+    this.renderJoinButton = this.renderJoinButton.bind(this)
+  }
+  renderMic() {
+    const { onLeftButtonPress, isInVoiceChannel } = this.props
+    return (
+      <TouchableOpacity onPress={onLeftButtonPress}>
+        <View style={styles.micView}>
```
EphemeralX: do you need this view? can the style just be in the TouchableOpacity?

civilapp/src/components/chat/MessageInput.js
```diff
@@ -45,34 +56,41 @@ class MessageInput extends Component {
...
         </View>
       </View>
-    ) : (
+    )
+  }
+
+  renderJoinButton() {
+    const { roomTitle, onPressMemberButton } = this.props
+    return (
       <TouchableOpacity onPress={() => onPressMemberButton()}>
```
EphemeralX: call the function directly rather than through an arrow function def

civilapp/src/components/chat/MessageInput.js
```diff
@@ -11,23 +11,34 @@ import SendButton from '../../../assets/images/chat/send_btn.png'
...
-      </View>
-      <View style={styles.micDropShadow} />
-    </TouchableOpacity>
-  )
+  constructor(props) {
+    super(props)
+    this.renderMic = this.renderMic.bind(this)
+    this.renderInputButton = this.renderInputButton.bind(this)
+    this.renderTextInput = this.renderTextInput.bind(this)
+    this.renderJoinButton = this.renderJoinButton.bind(this)
```
EphemeralX: are you sure you need to bind this?

rauchp: how else will I access the states from within the function?

civilapp/src/components/chat/MessageInput.js
rauchp: great catch, wasn't even looking at my oppresses, thanks!

civilapp/src/components/chat/MessageInput.js
rauchp: holy shit u boomed me

civilapp/src/components/chat/MessageInput.js
EphemeralX: oh I guess if you need it to access state, that's fine

PR #181, opened 2020-01-05
Optimized Profile stuff

rauchp:
- Made ProfileView fast.
- Made Edit ProfileView faster.
- Made Create Room Faster, added loading.
- Resolved Alex's nits for Profile.


civilapp/src/browse/SelectBannerScreen.js
```diff
@@ -102,11 +101,28 @@ class SelectBannerScreen extends Component {
...
+        <Text style={styles.text}>Making Your Room...</Text>
+      </View>
+    </>
+  )
   render() {
     return (
       <RetroBackground>
-        {this.renderHeader()}
-        {this.renderBanners()}
+        {this.state.loadingRoomInit ? (
```
EphemeralX: nit: can we get into the habit of destructuring state/props?  i.e. `const { loading } = this.state`

civilapp/src/profile/ProfileEditScreen.js
```diff
@@ -54,7 +54,7 @@ class ProfileEditScreen extends React.Component {
   initScreen = async () => {
     const { navigation } = this.props
     const userId = navigation.getParam('userId')
-    const userInfo = await UserAPI.getUser(userId)
+    const userInfo = navigation.getParam('userInfo')
```
EphemeralX: I think this was there originally to ensure that the user info is always up-to-date, do you foresee this being a problem?

civilapp/src/profile/ProfileScreenNew.js
```diff
@@ -181,7 +64,9 @@ class ProfileScreenNew extends React.Component {
       userId = userIdFromParams
       canEdit = false
     }
-    const userInfo = await UserAPI.getUser(userIdFromParams)
+
+    if (!canEdit) userInfo = await UserAPI.getUser(userId)
```
EphemeralX: nit: you guys may disagree but I find that even for 1 line if-statements, wrapping in `{}` is cleaner and more readable - I always do this for the code I touch/see.

civilapp/src/profile/ProfileEditScreen.js
rauchp: My thinking - the only way to edit your profile is by getting past this screen. Which means you can't ever be passed stale data, (assuming we reload the userInfo that's passed in, I'm doing those hooks in my current PR now.)

civilapp/src/profile/ProfileScreenNew.js
rauchp: nope I agree, I just didn't catch this.

civilapp/src/browse/CreateRoomScreen.js
```diff
@@ -83,7 +80,7 @@ export default class CreateRoomScreen extends React.Component {
         leftButtonPress={this.backButtonPress}
         rightButtonPress={() => {
           navigation.navigate('SelectBanner', {
-            createRoom: this.initRoom.bind(this)
+            createRoom: (banner) => this.initRoom(banner)
```
carteryu: is this better for performance? j/w

rauchp: nope, I was testing this out and forgot 2 change. turns out that binding wasn't ever necessary, and with `createRoom: this.initRoom` we get the same results as before and can even parameter pass still.

PR #182, opened 2020-01-05
Sportstab / Highlights / Profile Optimizations.

rauchp:
Don't want this PR to get too complicated, so I'm pushing now.

Done:
- Optimized highlights tab to make it load faster.
- Optimized SportsTab for fast loading
- Changed some Room Details Drawer things.
- Made Discovery and Browse less stale and more responsive. (Deleting/Unfollowing a room makes it dissapear when popped back to browse, same for joining a room.)
- (!) Redux'd the Profile Nav + Drawer. Now we don't have to rely on Async Storage for it all, (but still used as a backup just in case.)

To Do in next PR:
- HighlightsTab when game not set (logic for setting is in this PR.) 
- Make CreateRoomScreen faster by combining GraphQL calls into one custom resolver.


civilapp/src/browse/RoomsView.js
```diff
@@ -50,12 +68,10 @@ class RoomsView extends Component {
     this.willFocusSubscription.remove()
   }
 
-
-  static getDerivedStateFromProps(props, _) {
-    if(props.preloadedRooms)
-      return { rooms: props.preloadedRooms}
-    return {}
+  componentDidUpdate(_, prevState) {
```
carteryu: why is componentdidupdate better in this context?

rauchp:
Yep I just realized I didn’t show that, which sucks because I worked hard and making it switch instantly. What about yours? I was facing a 1 second delay when I tried something like u did

civilapp/src/browse/RoomsView.js
rauchp: getDerived is better when props are consistently changing, since it only changes once on load, didUpdate works better here.

civilapp/src/current/SportsTab.js
```diff
@@ -107,19 +108,21 @@ class SportsTab extends React.Component {
...
-    this.pullGameDataTick(gameId)
+    await this.pullGameDataTick(gameId)
     this.interval = setInterval(() => this.pullGameDataTick(gameId), 25000)
   }
 
   fetchGameInfo = async () => {
-    const roomInfo = await RoomAPI.getRoom(this.props.roomId)
-    const { gameId } = roomInfo
+    const { room } = this.props
+    let gameId = room.gameId ? room.gameId : ''
```
carteryu: can you make sure that when you hit a room with no game id that the no game selected view loads?

asking cuz idk if '' is the same as null

rauchp: m8 just look at the GIF. 

In JS, {} is the same as [] is the same as ‘’ is the same as undefined when it comes to null checking (!variable)

carteryu: literally just saw it and came back to close this lol.

mb

civilapp/src/browse/RoomsView.js
rauchp: what lifecycle would you prefer? Since we're using a promise, preloadedRooms might not be ready on mount or in the constructor.  

Checking for that conditional is low-cost in didUpdate is low cost, and it'll only run a max of one time.

civilapp/src/browse/BrowseScreen.js
```diff
@@ -131,7 +142,9 @@ class BrowseScreen extends Component {
     return (
       <RetroBackground>
         {this.renderHeader()}
-        <RoomsView preloadedRooms={preloadedRooms} />
+        <RoomsView
+        preloadedRooms={preloadedRooms}
```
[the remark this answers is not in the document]
rauchp: think I just'd it

PR #185, opened 2020-01-07
UI Nits

EphemeralX:
* UI nits for RoomView
* UI nits PlayerStatCard
* Refactored out NoGameSelectedView
* Add NoGameSelectedView to HighlightsTab
* Update BannerUtil to have 2 options (day and night version of Staple Center))
* Browse/Discover/Social header nit fixes
* Add Video to GradientDropShadow
* Add silent repeating video (fake gif) to Highlights and Stats UI when no game is selected
* Update night mode staples center banner asset


civilapp/src/browse/RoomsView.js
```diff
@@ -69,7 +68,11 @@ class RoomsView extends Component {
   }
 
   componentDidUpdate(_, prevState) {
-    if (prevState.rooms.length === 0 && this.props.preloadedRooms) this.setState({ rooms: this.props.preloadedRooms })
+    const { preloadedRooms } = this.props
+    if (prevState.rooms.length === 0 && preloadedRooms) {
+      this.setState({ rooms: preloadedRooms })
+      setTimeout(() => {}, 1000)
```
rauchp: why?

civilapp/src/components/WoodenBackground.js
```diff
@@ -57,7 +57,7 @@ const styles = {
     paddingTop: 30
   },
   alertView: {
-    flexBasis: 64,
+    flexBasis: 24,
```
rauchp: why? sigma had height of 64 I thought

civilapp/src/browse/RoomsView.js
EphemeralX: whoops, this is still causing an infinite setState loop on android - I tried to see if this would fix it but it didn't, will remove.

civilapp/src/components/WoodenBackground.js
EphemeralX: figma included the safe area height, without it, it's 24

rauchp: why flex basis over height?

EphemeralX: it's technically same thing but only works with flex items.  I was experimenting with this when trying to fix the safe area thing and never changed it back and it was a 'not broken, don't fix it' moment

PR #186, opened 2020-01-07
CreateRoom SpeedUp + SportsTab Fix + Discover AddRoom Consistency 

rauchp:
- Fixed SportsTab Updating
- Added custom resolver to make CreateRoom faster
- Made new RoomAPI call + lambda.
- Made DiscoverScreen's addRoom accurate.
- Some nits here and there.


civilapp/src/browse/CreateRoomScreen.js
```diff
@@ -23,15 +23,14 @@ export default class CreateRoomScreen extends React.Component {
 
     this.initSubscribers = navigation.getParam('initSubscribers', '')
     const userInfo = navigation.getParam('userInfo', '')
-    this.addRoom = navigation.getParam('addRoom', (room) => {})
+    this.addRoom = navigation.getParam('addRoom', () => {})
```
carteryu: havent seen this before, why is empty arrow func needed? o.o

civilapp/src/current/GameSelectScreen.js
```diff
@@ -59,27 +60,27 @@ class GameSelectScreen extends React.Component {
...
+          this.updateGameId(navigation.getParam('roomId'), selectedGameId)
         }}
-        middleText={'SELECT GAME'}
-        leftText={'<<'}
-        rightText={'SAVE'}
-        disabled={selectedGameId === 'none' ? true : false}
+        middleText='SELECT GAME'
+        leftText='<<'
+        rightText='SAVE'
+        disabled={selectedGameId === 'none'}
```
carteryu: reee.. i did this too lol

civilapp/src/browse/CreateRoomScreen.js
rauchp: so we don't get a weird error by trying to call a function on a {} or '' object.

PR #188, opened 2020-01-07
FTUE tweaks, replace AsyncStorage USER_INFO calls, removed upheaval

EphemeralX:
* Replace `AsyncStorage.getItem('USER_INFO')` with `getUserInfo` from `AuthUtils` - this check cognito/dynamoDB and isn't as error prone.
* Update FTUE pixels for phone number input sizing
* Add saving indicators for next header in FTUE
* PureComponents and less inline arrow functions defined in renders
* ESlinter BS
* Keyboard padding fix in RoomView after adjusted Green Alert Mic On component
* Removed unused styles and imports in GameSelectScreen
* Fixed borderWidth to 1 from 1.5 as in designs
* Removed some upheaval references


civilapp/src/auth/ConfirmSignInScreen.js
```diff
@@ -96,11 +89,13 @@ export default class ConfirmSignInScreen extends Component {
     const authenticated = await this.confirmAuthSession()
     console.log('Authenticated:', authenticated)
     if (!authenticated) {
+      this.setState({ saving: false })
       return
     }
     const { userId, isNewUser } = this.state
     const nextRoute = isNewUser ? 'UsernameSetup' : 'Loading'
     const { navigation } = this.props
+    this.setState({ saving: false })
```
rauchp: I am 99% when we do this instead of 

this.setState({saving : false}, () => {
navigation.navigate....
})

we cause the called setState on an unmounted component.
Since setState is async, there are some times where it'll navigate before setState started and we start setStating on the wrong thing.

civilapp/src/browse/RoomsView.js
```diff
@@ -43,7 +43,10 @@ class RoomsView extends Component {
 
   addRoom = room => {
     const newRooms = this.state.rooms
-    newRooms.unshift(room)
+    // We need to check if the room doesn't already exist in the list in case it's the default subscribed room new users are created with
```
rauchp: wait what? addRooms is for rooms you join or create yourself, if we were already sub'd to a room, you wouldn't be able to add it?

civilapp/src/current/SportsTab.js
```diff
@@ -520,7 +520,7 @@ class SportsTab extends React.Component {
 
         <View style={[styles.centeredView, { marginHorizontal: LeftwardMargin + 1.5 }]}>
           <ScrollableTabView
-            style={isIOS ? {} : { width: '100%' }}
+            style={Platform.OS === 'ios' ? {} : { width: '100%' }}
```
rauchp: in my pr I just added isIOS back to top of file (not sure how it ever disappeared?), are we gonna have an unused isIOS const after merge

civilapp/src/profile/ProfileScreenNew.js
```diff
@@ -0,0 +1,274 @@
+import React from 'react'
```
rauchp: I thought Carter removed this file in the one of his merged PR's, are we bringing it back?

civilapp/src/auth/ConfirmSignInScreen.js
EphemeralX: this is what I told carter but it hasn't happened to him when I asked (wouldn't surprise me if navigate was slower than setState) so I figured I'd be consistent, but totally down to change it all to be consistent in the other direction though

civilapp/src/browse/RoomsView.js
EphemeralX: there is an edge case when you create a new account, you see Rush Team and when you press it, you still have to join it, and if you do, it duplicates that entry in the code.  This is to prevent that duplication from happening.

civilapp/src/current/SportsTab.js
EphemeralX: we can use it

civilapp/src/profile/ProfileScreenNew.js
EphemeralX: whoops, will remove

civilapp/src/browse/RoomsView.js
rauchp: Ah ok gotcha. But that only happens because we’re also not making them a member of the auto joined rooms on user init. We can just make sure we’re adding them as members too, or have this code. Actually don’t mind bc this code isn’t asynch so it’s still performant

EphemeralX: actually it shouldn't be an issue with your new lambda

rauchp:
where are we setting them to join rush team?

civilapp/src/browse/RoomsView.js
EphemeralX: nvm, got the code mixed up - could be fixed the same though but for now it's okay

PR #189, opened 2020-01-07
Small PR. Fixed depth, and null on delete room.

rauchp:
- Fixed some lint things
- Fixed infinite setState
- Fixed error when getting from BannerUtil, added error check to map get
- Changed refresh for DiscoveryRoomsView to match BrowseScreen.
- Fixed null error when deleting a room (it was set to willBlur instead of willFocus?)


civilapp/src/browse/RoomsView.js
```diff
@@ -69,7 +69,7 @@ class RoomsView extends Component {
 
   componentDidUpdate(_, prevState) {
     const { preloadedRooms } = this.props
-    if (prevState.rooms.length === 0 && preloadedRooms) {
+    if (prevState.rooms.length === 0 && preloadedRooms.length > 0) {
```
EphemeralX: does this fix the infinite loop on android?

civilapp/src/current/EditRoomDetails.js
```diff
@@ -89,7 +89,7 @@ class EditRoomDetails extends React.Component {
   componentDidMount = async () => {
     this.initScreen()
     const { navigation } = this.props
-    this.willBlurSubscription = navigation.addListener('willBlur', payload => {
+    this.willBlurSubscription = navigation.addListener('willFocus', payload => {
```
EphemeralX: this is still called willBlurSubscription

civilapp/src/browse/RoomsView.js
rauchp: Yep

PR #193, opened 2020-01-08
Janky input fix

rauchp:
- fixed ftue janky input on last key pressed


civilapp/src/auth/SignInScreen.js
```diff
@@ -231,19 +230,15 @@ export default class SignInScreen extends PureComponent {
...
-    this.showLoading(false)
-  }
-
-  renderHeader = () => {
-    const { enableButton, saving } = this.state
+  renderHeader() {
+    const { saving } = this.state
     return (
-      <SuperTextHeader
+      <InputSuperTextHeader
```
EphemeralX: is this specific to the FTUE in any way?

rauchp:
only in that we’ll never use this outside of these two screens.

PR #196, opened 2020-01-08
Disable share on non member room + eslint nits

carteryu:


civilapp/src/current/SportsTab.js
```diff
@@ -320,12 +329,13 @@ class SportsTab extends React.Component {
...
 
   renderModal = () => {
+    const { modalVisible } = this.state
     return (
       <RetroModal
         hasKeyboard
         iosOffset={-80}
-        toggleModal={() => this.setModalVisible(!this.state.modalVisible)}
-        isVisible={this.state.modalVisible}
+        toggleModal={() => this.setModalVisible(!modalVisible)}
```
EphemeralX: remember to factor this out into a normal function since it check state

PR #199, opened 2020-01-09
headshots + roomdetails fixes

carteryu:
game select slides from right
changed how roomdetails refreshes data
new player headhots

might be some delay with loading playercard since it uses await to fetch from s3. if its a problem on TF ill change it


civilapp/src/sports/utils/FlairUtil.js
```diff
@@ -154,112 +153,111 @@ const EasternConferenceData = [
...
     primaryColor: '#002B5C',
     secondaryColor: '#E31837',
-    altColor: '#C4CED4',
-  },
+    altColor: '#C4CED4'
+  }
 ]
 
-let FlairDict = convertArrayToObject([...WesternConferenceData, ...EasternConferenceData, AnonymousObject], 'key')
+const FlairDict = convertArrayToObject([...WesternConferenceData, ...EasternConferenceData, AnonymousObject], 'key')
```
carteryu: only thing i changed in this file.

civilapp/src/current/common/PlayerStatCard.js
```diff
@@ -26,7 +29,8 @@ class PlayerStatCard extends Component {
     super(props)
     this.playerInfo = this.props.playerInfo
     this.state = {
-      loading: true
+      loading: true,
+      bgColor: 'transparent'
```
EphemeralX: why is this a state?

civilapp/src/sports/utils/FlairUtil.js
```diff
@@ -30,8 +30,7 @@ import UtahJazz from '../../../assets/logos/NBA/Utah_Jazz_logo.png'
 import WashingtonWizards from '../../../assets/logos/NBA/Washington_Wizards_logo.png'
 import AnonymousUser from '../../../assets/avatars/default_profile.png'
 
-const convertArrayToObject = (array, key) =>
-  array.reduce((obj, item) => ((obj[[item[key]]] = item), obj), {})
+const convertArrayToObject = (array, key) => array.reduce((obj, item) => ((obj[[item[key]]] = item), obj), {})
```
EphemeralX: you can probably use Object.entries?

civilapp/src/current/RoomDetails.js
```diff
@@ -93,15 +95,38 @@ class RoomDetails extends React.Component {
...
+      const gameTime = gameObj ? this.convertToLocalTime(gameObj.DateTime) : null
+      this.setState({
+        homeTeam,
+        awayTeam,
+        gameStatus,
+        gameDay,
+        gameTime,
+        gameSelected: true
+      })
+    } else {
```
EphemeralX: jw, instead of setting these as individual fields, why not just set these as 1 gameObj? then just access the fields on the gameObj when you need them.

civilapp/src/current/RoomDetails.js
```diff
@@ -60,16 +61,17 @@ class RoomDetails extends React.Component {
...
-      roomInfo: this.fetchRoomInfo()
-    })
     let userId = navigation.getParam('userId')
     if (!userId) {
       const authenticatedUserInfo = await getUserInfo()
       userId = authenticatedUserInfo.id
     }
-    this.setState({ userId })
+    this.setState({ userId, gameId: roomInfo.gameId, chatChannelUrl: roomInfo.chatChannelUrl, room: roomInfo })
+    this.willFocusSubscription = navigation.addListener('didFocus', () => {
```
EphemeralX: this is called `willFocus` but is actually `didFocus`

civilapp/src/current/common/PlayerStatCard.js
carteryu: i copied the s3 fetch code directly from your landing transition

bgcolor state is just the default color (which should never happen)

civilapp/src/sports/utils/FlairUtil.js
carteryu: this is pedros code, didnt wanna touch in case it broke other parts of the codebase its hard for me to debug

civilapp/src/current/GameSelectScreen.js
```diff
@@ -37,9 +37,10 @@ class GameSelectScreen extends React.Component {
 
   componentDidMount = async () => {
     const { navigation } = this.props
-    this.willFocusSubscription = navigation.addListener('didFocus', () => {
-      this.fetchGameInfo()
-    })
+    this.fetchGameInfo()
+    // this.willFocusSubscription = navigation.addListener('didFocus', () => {
+    //   this.fetchGameInfo()
+    // })
```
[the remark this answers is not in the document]
carteryu: its gone in next pr, my bad

civilapp/src/current/CurrentlyPlayingScreen.js
```diff
@@ -125,7 +125,9 @@ export default class CurrentlyPlayingScreen extends Component {
 
   navigateToGameSelect = () => {
     const { navigation } = this.props
-    navigation.navigate('GameSelect', { isDefault: true })
+    const { room } = this.state
+    const roomId = room.id
```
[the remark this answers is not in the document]
carteryu: just so i can shorthand roomId param, i can rm just reply to this comment

civilapp/src/current/RoomDetails.js
carteryu: 1. cuz i dont want to do those null checks in the render method
2. cuz sometimes i only want to update one of the 5 states. i wouldnt want to recreate a new game object and dissect it each time

civilapp/src/current/RoomDetails.js
carteryu: fixing

PR #202, opened 2020-01-10
Newerer roomview

rauchp:
- turns out agora init was broken, fixed it and a new world of possibilities opened up
- now userJoin only notifies us for broadcast joins, which is exactly what I wanted
- cleaned up RoomView, and listened to linter:)
- shit works way faster now.

wg5P7W2


civilapp/src/current/RoomView.js
```diff
@@ -406,85 +373,48 @@ class RoomView extends React.Component {
...
-
-    const addRoom = this.props.navigation.getParam('addRoom')
-    if (addRoom) {
-      addRoom(room)
+      await RoomAPI.removeMutedUser(room.id, userInfo.id)
     }
-    this.setState({ isMember: true })
+    setInVoiceChannel(!isInVoiceChannel)
+    this.isInVoiceChannel = !isInVoiceChannel
+    this.setState({})
```
EphemeralX: if your intention is to re-render, you can just call this.forceUpdate()

rauchp:
basically deprecated the need for speaker list, which is YUGE b/c that's a shit ton less async/await API calls. Besides this, I quit relying on unreliable DynamoDB subscriptions for client side speaker list settings, and switched to (ironically) more reliable Agora Subscriptions. This also means I was able to remove the SendBird user list calls I was using to confirm speaker list.

rauchp:
also for some reason it's hella fast on initial room setup on sim, but slow on device. But for both it's fast afterwards. This might be a product of my Bosnian wifi tho, can't tell. Also might be faster on TF in release mode

PR #203, opened 2020-01-11
newererer roomview fix 

rauchp:
- relationship with debug mode over, release mode my new best friend 

fuckers in the app store telling me, always in testflight
"pedro ain't 'bout this, roomview ain't 'bout that
My app slow as shit on fucking Lamron and them
He, he, they say that ***** don't be putting in no work”
Shut the fuck up!
Y'all **** ain’t know shit!
All ya motherfuckers talk about...
“roomview ain't as fast it used to be, pedro dropping bad pr's, rush aint even good”
Shut the fuck up!
Y'all don't real with that *****!


civilapp/src/auth/AuthLoadingScreen.js
```diff
@@ -58,6 +59,7 @@ export default class AuthLoadingScreen extends PureComponent {
     const { navigation } = this.props
     try {
       const userId = await initializeUserData()
+      AgoraAPI.enterListenerChannel('123', userId).then(() => AgoraAPI.leaveChannel())
```
EphemeralX: what's this parameter?

civilapp/src/current/RoomView.js
```diff
@@ -51,21 +42,51 @@ class RoomView extends React.Component {
...
+    }
   }
 
   componentWillUnmount() {
     this.clearRoom()
   }
 
+  initializeRoom() {
+    const { userInfo, roomInitialSetup, setInVoiceChannel, navigation } = this.props
+    const { room } = this.props
```
EphemeralX: room can be on the previous line

civilapp/src/auth/AuthLoadingScreen.js
rauchp: ‘’ doesn’t work because it’s treated as null 

basically, we need to enter a channel before the first roomview launch because it lags the app by 1-1.5 seconds. We should do this in loading transition ideally, but I guess we can move it if we notice the lag in authLoading

PR #204, opened 2020-01-12
Nits prelaunch

rauchp:
- Muted video when you leave the highlights tab.
- Footer blinks now when you slide across to another page.
- Fixed this: "Changing title in edit room doesnt auto update room title "
   - fixed for both currentlyPlayingScreen header and roomDetails
- fixed deletingRooms so your own room disappears when you delete.
- fixed deleted RoomView so it actually works now.


civilapp/src/components/CustomFooter.js
```diff
@@ -17,42 +17,41 @@ const RedSelector = require('../../assets/images/footer/red_selector.png')
...
+    ) : (
+      <Text style={styles.footerText}>{props.text}</Text>
+    )}
+  </TouchableOpacity>
+))
 
 class CustomFooter extends Component {
   constructor(props) {
     super(props)
+    this.buttons = [0, 0, 0]
```
rauchp: open to suggestions for a cleaner way of doing this

civilapp/src/current/RoomDetails.js
```diff
@@ -61,22 +51,24 @@ class RoomDetails extends React.Component {
...
     let userId = navigation.getParam('userId')
     if (!userId) {
       const authenticatedUserInfo = await getUserInfo()
       userId = authenticatedUserInfo.id
     }
+
     this.setState({ userId, chatChannelUrl: roomInfo.chatChannelUrl, room: roomInfo })
-    this.didFocusSubscription = navigation.addListener('didFocus', () => {
-      this.updateGameData()
+    this.willFocusSubscription = navigation.addListener('willFocus', () => {
```
carteryu: hey whats diff between will and did focus in this context?

all i know is that didfocus only triggers after coming back from gameselect or editroom so in that case you only care about updating the gamedata/bio and not everytthing else. u sure this is good?

rauchp:
will focus === the screen is about to appear

did focus === the screen appeared

with didfocus, you’ll notice some lag, with willFocus it’ll activate when
you start navigating not when it’s already done

civilapp/src/current/RoomViewError.js
```diff
@@ -1,68 +1,69 @@
...
-} from 'react-native'
+import { TouchableOpacity, Dimensions, StyleSheet, View, Platform } from 'react-native'
 import { withNavigation } from 'react-navigation'
 import { Text } from 'native-base'
 import RetroBackground from '../components/RetroBackground'
-const { width, height } = Dimensions.get('window')
+
+const { height, width } = Dimensions.get('window')
 
 const UPHEAVAL = Platform.OS === 'ios' ? 'Upheaval TT (BRK)' : 'upheavtt'
```
EphemeralX: not a big priority but we're trying to get rid of upheaval entirely in favor of sportsball (but that'd require a font size change as wel)

civilapp/src/current/CurrentlyPlayingScreen.js
```diff
@@ -168,30 +190,18 @@ export default class CurrentlyPlayingScreen extends Component {
...
-          renderTabBar={() => (
-            <CustomRoomTabBar
-              contentContainerStyle={{
-                marginTop: 3,
-                marginBottom: -5
-              }}
-              overlayBottom
-            />
-          )}
+          renderTabBar={() => <CustomRoomTabBar contentContainerStyle={styles.tabContainer} overlayBottom />}
```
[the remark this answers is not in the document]
rauchp: I didn't write this code originally :(

PR #211, opened 2020-01-17
Rush dfs user info flow

rauchp:
Finished UI for user info flow. Need to wire up with Amplify calls. Marked with TODO's where there's work to be done for data logic.

Simulator Screen Shot - iPhone 11 - 2020-01-16 at 19 30 08
Simulator Screen Shot - iPhone 11 - 2020-01-16 at 19 30 31
Simulator Screen Shot - iPhone 11 - 2020-01-16 at 19 30 27
Simulator Screen Shot - iPhone 11 - 2020-01-16 at 19 30 20
Simulator Screen Shot - iPhone 11 - 2020-01-16 at 19 30 15
Simulator Screen Shot - iPhone 11 - 2020-01-16 at 19 30 12
Simulator Screen Shot - iPhone 11 - 2020-01-16 at 19 30 09


rauchp:
planing on re-using most of the amplify logic from rush chat, just need to merge carter's stuff before I can start making calls...

RushDFS/src/auth/common/Dots.js
```diff
@@ -0,0 +1,27 @@
...
+          marginHorizontal: 1.75,
+          backgroundColor: index < activeDots ? '#FDB927' : '#151E25'
+        }}
+      />
+    )
+  }
+
+  render() {
+    const { totalDots } = this.props
+    return <View style={{ flexDirection: 'row' }}>{Array.from(Array(totalDots)).map(this.renderDot.bind(this))}</View>
```
EphemeralX: can we put this Array logic in componentDidMount or constructor so it doesn't execute every render?

RushDFS/src/auth/UsernameInputScreen/index.js
```diff
@@ -0,0 +1,278 @@
...
+    const regex = /[ !@#“‘øßłœ”’$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]/g
+
+    if (!regex.test(text) && text.length <= AUTH_USERNAME_MAX_LENGTH) {
+      clearTimeout(this.timer)
+      this.timer = null
+      this.testUsername(text)
+    }
+  }
+
+  testUsername = username => {
```
EphemeralX: nit: I think to shorten this function, we could factor our the success and error cases into their own functions and call them here.

RushDFS/src/auth/InitialNumberInputScreen/index.js
```diff
@@ -0,0 +1,220 @@
...
+      </View>
+    )
+  }
+
+  renderButton() {
+    const { navigation } = this.props
+    const { number } = this.state
+    const disabled = number.length < AUTH_NUMBER_LENGTH
+
+    let color = this.state.bgColor.interpolate({
```
EphemeralX: const

EphemeralX: deconstruct bgColor

RushDFS/src/auth/InitialNumberInputScreen/index.js
```diff
@@ -0,0 +1,220 @@
...
+    return <AuthHeader onPress={() => navigation.goBack()} activeDots={1} totalDots={isSignIn ? 2 : 5} />
+  }
+
+  renderTitle = (<Text style={styles.title}>Phone Number</Text>)
+
+  renderSubtitle = (
+    <Text style={styles.subtitle}>We use your phone number for account creation and verification purposes only</Text>
+  )
+
+  changeText = text => {
```
EphemeralX: optional, but I wonder if you could simplify this logic

RushDFS/src/auth/InitialNumberInputScreen/index.js
```diff
@@ -0,0 +1,220 @@
...
+      Animated.timing(this.state.bgColor, { toValue: 1, duration: 325 }).start()
+    }
+  }
+
+  renderInput() {
+    const { number } = this.state
+    const finished = number.length === AUTH_NUMBER_LENGTH
+    return (
+      <View style={styles.inputView}>
+        <TextInput
```
EphemeralX: notice you don't have autoCorrect and autoCapitalize set to off/false here like you do in the other TextInputs, intentional?

RushDFS/src/auth/EmailInputScreen/index.js
```diff
@@ -0,0 +1,273 @@
...
+
+    Animated.timing(bgColor, { toValue: 0, duration: 400 }).start()
+    this.setState({ secondaryMessage: ' ', secondaryColor: 'white' }, () => {
+      // TODO: add call or better huesteric to determine if valid email.
+      const isValid =
+        email &&
+        email.includes('.') &&
+        email.includes('@') &&
+        email[email.length - 1] !== '@' &&
+        email[email.length - 1] !== '.'
```
EphemeralX: I think there are email validation regexes online or something you can use

RushDFS/src/auth/EmailInputScreen/index.js
```diff
@@ -0,0 +1,273 @@
...
+    const regex = /[ !#“‘øßłœ”’$%^&*()_+\-=\[\]{};':"\\|,<>\/?]/g
+
+    if (!regex.test(text)) {
+      clearTimeout(this.timer)
+      this.timer = null
+      this.testEmail(text)
+    }
+  }
+
+  testEmail = email => {
```
EphemeralX: nit: this and the other testSomething method below imo should be renamed to something like validateEmail, validateUsername or whatever.  test is kind of... odd imo

EphemeralX: nit: also try to aim for shorter functions; the valid and invalid cases below could be factored out into it's own functions to make this one shorter.

RushDFS/src/auth/EmailInputScreen/index.js
```diff
@@ -0,0 +1,273 @@
...
+import { SafeAreaView } from 'react-native-safe-area-context'
+
+import AuthHeader from '../common/AuthHeader'
+import { BACKGROUND_BLACK, AUTH_LEFT_ALIGN, AUTH_DIST_FROM_TOP, GREEN_SUCCESS, RED_ERROR } from '../../Consts'
+
+const { height } = Dimensions.get('window')
+
+const isSmallScreen = height < 600
+const isIOS = Platform.OS === 'ios'
+const ROOM_TILE = 'Email'
```
EphemeralX: haha found it

RushDFS/src/auth/InitialNumberInputScreen/index.js
rauchp: it's numeric so there's really no option here

RushDFS/src/auth/common/Dots.js
rauchp: actually, because it's a PureComponent, it's not re-rendering on every render (since no state or prop changes between the rendering of <Dot /> itself)

